PR #16, opened 2021-05-26
Preserve Gmail Privacy

edent:
Users have legitimate reasons for adding tracking information to their email addresses. To preserve users' security and privacy, these tracking codes should be maintained.

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -51,13 +51,6 @@ When sending email addresses in a request, normalize email addresses prior to se
 1. Remove leading and trailing spaces.
 2. Convert all ASCII characters to lowercase.
 
-For email accounts ending in gmail.com:
-
-1. Remove leading and trailing spaces.
-2. Convert all ASCII characters to lowercase.
-3. Remove `.` (ASCII code 46) from the username of the email address. `jane.doe@gmail.com` normalizes to `janedoe@gmail.com`.
-4. Remove `+` (ASCII code 43) and all subsequent characters from the username of the email address. `janedoe+home@gmail.com` normalizes to `janedoe@gmail.com`.
-
 ## Encoding Query Parameter Values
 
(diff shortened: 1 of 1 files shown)


atarassov-ttd:
Thank you for your input, we thought long about this update and ultimately as it stands today it is not a change we would like to add. We appreciate your feedback and hope you continually help to make enhancements.

edent:
Why is it OK for IAB members to track users - but it isn't OK for users to track IAB members?

brouhaha:
This is terrible. There are also non-tracking reasons to use a +suffix in an email address. That's a perfectly legitimate email address according to RFCs, and should not be arbitrarily mangled, whether the address is associated with gmail or not.

sxpert:
as a reminder, GDPR fines start at 20M€... and can go up to 4% worldwide yearly revenue

EmSixTeen:
> Thank you for your input, we thought long about this update and ultimately as it stands today it is not a change we would like to add. We appreciate your feedback and hope you continually help to make enhancements.

Why do you believe you should get the right to edit the details someone has specifically entered when signing up? You can't change their name. You can't change their address. Why do you believe you can change their email address? Please share your justification.

TedTed:
Hello! Data accuracy is the fourth GDPR [Principle of Data Protection](https://www.dataprotection.ie/en/individuals/data-protection-basics/principles-data-protection). The ICO (Data Protection Authority for the UK) [guidance](https://ico.org.uk/for-organisations/guide-to-data-protection/guide-to-the-general-data-protection-regulation-gdpr/principles/accuracy/#accurate_inaccurate) states that "You must always be clear about what you intend the record of the personal data to show".

"Normalizing" email addresses, and silently transform e.g. johndoe+company@gmail.com into johndoe@gmail.com, would fairly clearly contravene this principle, as this user would have no way of knowing that their data record was altered in this way.

What is your proposed solution to this fairly obvious compliance issue?

PR #22, opened 2021-07-15
Add language selector

miyaichi:
A simple language selector has been added in anticipation of this document being translated into multiple languages in the future.


vlochner (commented):
Can we spell out the full name of each language please, since there is currently no concern about space?

miyaichi:
Fixed to spell out the full name in each language.

PR #42, opened 2022-07-18
Correct callback property name

dlackty:
Accoding to the [SDK source code](https://prod.uidapi.com/static/js/uid2-sdk-2.0.0.js):

```js
const updateStatus = (status, statusText) => {
    _lastStatus = status;

    const promises = _promises;
    _promises = [];

    const advertisingToken = this.getAdvertisingToken();

    const result = {
        advertising_token: advertisingToken,
        status: status,
        statusText: statusText
    };
    _opts.callback(result);

    if (advertisingToken) {
        promises.forEach(p => p.resolve(advertisingToken));
    } else {
        promises.forEach(p => p.reject(new Error(statusText)));
    }
};
```

The correct property name should be `advertising_token` instead of `advertisingToken`.

### DIFF
--- a/api/v2/sdks/client-side-identity.md
+++ b/api/v2/sdks/client-side-identity.md
@@ -165,7 +165,7 @@ The `object` parameter includes the following properties.
 
 | Property | Data Type | Description |
 | :--- | :--- | :--- |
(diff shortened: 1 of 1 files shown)


elenaTTD:
The SDK will be updated, not the docs.

dlackty:
@elenaTTD I just sent a pull request to the SDK repo. Please have a look. Thanks.

PR #149, opened 2023-04-19
improve python examples

jon8787:


genwhittTTD (approved):

genwhittTTD (approved):
Looks good. Also I verified locally that the coding passes validation on the Docusaurus build.

genwhittTTD (approved):
There is no request example at all, now, for POST token/refresh. Is that as expected? There is a response example. I think we should have both. It links to the other doc, but that doc doesn't include a POST token/refresh example. I think we've lost something that we need in this file. Just commenting, not blocking -- your call.

PR #151, opened 2023-04-19
hashing encoding doc mods

genwhittTTD:
Updates to clarify function to use for hashing/encoding.


docs/getting-started/gs-normalization-encoding.md
```diff
@@ -43,13 +43,15 @@ To normalize an email address, complete the following steps:
 
 ## Email Address Hash Encoding
 
-An email hash is a Base64-encoded SHA-256 hash of a normalized email address. The email address is first normalized, then hashed using the SHA-256 hashing algorithm, and then encoded using Base64 encoding.
+An email hash is a Base64-encoded SHA-256 hash of a normalized email address. The email address is first normalized, then hashed using the SHA-256 hashing algorithm, and the resulting hex value is encoded using Base64 encoding.
```
atarassov-ttd, line 46: Should be something like:

> the resulting bytes of the hash value (rather than their hex-encoded string representation) is encoded using Base64 encoding

docs/getting-started/gs-normalization-encoding.md
```diff
@@ -43,13 +43,15 @@ To normalize an email address, complete the following steps:
...
 
 | Type | Example | Comments and Usage |
 | :--- | :--- | :--- |
 | Normalized email address | `user@example.com` | Normalization is always the first step. |
-| SHA-256 hash of normalized email address | `b4c9a289323b21a01c3e940f150eb9b8c542587f1abfd8f0e1cc1ffc5e475514` | This 64-character string is a hex-encoded representation of 32-byte SHA-256.|
-| Base64-encoded SHA-256 of normalized email address | `tMmiiTI7IaAcPpQPFQ65uMVCWH8av9jw4cwf/F5HVRQ=` | This 44-character string is a Base64-encoded representation of 32-byte SHA-256.<br/>Use this encoding for `email_hash` values sent in the request body. |
+| SHA-256 hash of normalized email address | `b4c9a289323b21a01c3e940f150eb9b8c542587f1abfd8f0e1cc1ffc5e475514` | This 64-character string is a hex-encoded representation of the 32-byte SHA-256.|
+| Hex to Base64 SHA-256 encoding of normalized email address | `tMmiiTI7IaAcPpQPFQ65uMVCWH8av9jw4cwf/F5HVRQ=` | This 44-character string is a Base64-encoded representation of the 32-byte SHA-256.<br/>NOTE: The SHA-256 hash is a hexadecimal value. You must use a Base64 encoder that takes a hex value as input.<br/>Use this encoding for `email_hash` values sent in the request body. |
```
atarassov-ttd, line 52: Should be something like:

> WARNING: The SHA-256 hash string in the example above is a hex-encoded representation of the hash value. You must Base64-encode the raw bytes of the hash or use a Base64 encoder that takes a hex-encoded value as input.

docs/getting-started/gs-normalization-encoding.md
```diff
@@ -43,13 +43,15 @@ To normalize an email address, complete the following steps:
...
+
+>WARNING: When applying Base64 encoding, be sure to use a function that takes a hex value as input. If you use a function that takes text as input, the result is a longer string which is invalid for the purposes of UID2.
```
atarassov-ttd, line 54: May want to update similar to the above.

atarassov-ttd (approved):

PR #174, opened 2023-05-28
docs(glossary-uid.md): minor edits

gaga5lala:
Thanks for providing a comprehensive document. While reading through the document, I identified these small areas where adjustments can enhance readability and maintain consistency.

1. I noticed that there is a missing opening parenthesis for JWT
2. I suggest changing "Opt-out" to "Opt-Out" for consistency in capitalization.


genwhittTTD:
@gaga5lala thank you so much! Made the first mod in another PR and merged.

On Opt-out: it's correct as-is since in the glossary we are using sentence capitalization (initial cap on the first word only) other than acronyms and product names. JSON Web Token and some other terms are in that category, so there is a mix in the capitalization of the glossary terms, but sentence capitalization is our basic guidance.

Very much appreciate your input.

PR #193, opened 2023-06-09
snowflake migration doc

genwhittTTD:
Added migration content in a new section. One marked query for MC.

### DIFF
--- a/docs/guides/snowflake_integration.md
+++ b/docs/guides/snowflake_integration.md
@@ -13,8 +13,8 @@ sidebar_position: 04
 - [Shared Objects](#shared-objects)
   -  [Database and Schema Names](#database-and-schema-names)
   -  [Map DII](#map-dii)
-  -  [Regenerated UID2s](#regenerate-uid2s) -->
-
+  -  [Regenerated UID2s](#regenerate-uid2s) 
+- [Migration Guide](#migration-guide)  -->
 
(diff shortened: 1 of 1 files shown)


docs/guides/snowflake_integration.md
```diff
@@ -431,3 +434,75 @@ The following table identifies each item in the response. The result includes an
...
+```
+FN_T_UID2_IDENTITY_MAP(EMAIL_HASH, 'email_hash')
+```
+
+### Using the Values for the UNMAPPED Column
+When you have the new function implemented, you can check the `UNMAPPED` column returned by the `FN_T_UID2_IDENTITY_MAP`. If any DII could not be mapped to a UIS2, this column gives the reason.
+
+For details about the values and their explanations, see [Values for the UNMAPPED Column](#values-for-the-unmapped-column).
+
+{**MC/GWH: we already have this info in the [Values for the UNMAPPED Column](#values-for-the-unmapped-column) section. Can I just point to that rather than having it twice? Please look -- I reworked the explanations in that section a little, and I added the detail about the values for `INVALID INPUT TYPE` into the table above.**}
```
mcollins-ttd, line 490: Yes, I agree it's better to have that information in a single place.

docs/guides/snowflake_integration.md
```diff
@@ -431,3 +434,75 @@ The following table identifies each item in the response. The result includes an
...
+```
+
+After:
+
+```
+FN_T_UID2_IDENTITY_MAP(EMAIL_HASH, 'email_hash')
+```
+
+### Using the Values for the UNMAPPED Column
+When you have the new function implemented, you can check the `UNMAPPED` column returned by the `FN_T_UID2_IDENTITY_MAP`. If any DII could not be mapped to a UIS2, this column gives the reason.
```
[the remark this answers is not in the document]
genwhittTTD: TYSM. Fixed.

docs/guides/snowflake_integration.md
genwhittTTD: TY. See update.

PR #203, opened 2023-06-28
mods to POST token/refresh per input from BL/SW

genwhittTTD:
mods to POST token/refresh per input from BL/SW. Queries marked (4).


docs/endpoints/post-token-refresh.md
```diff
@@ -19,9 +19,8 @@ Used by: This endpoint is used mainly by publishers.
 Here's what you need to know about this endpoint:
 
 - No encryption is required for token refresh requests.
-- Responses are encrypted only if the HTTP status code is 200. Otherwise, responses are not encrypted.
-- To decrypt responses, the `refresh_response_key` value returned in the [POST /token/generate](post-token-generate.md) or `POST /token/refresh` response from which the refresh token in the request is returned.
-- If you send a refresh token from a v1 `token/generate` response in the request, the response will not be encrypted.
+- If the request is successful, with an HTTP status code of 200 and a new UID2 token or opt-out information returned, the response is encrypted. Error responses are not encrypted. [**GWH/SW_01 opt-out is a 200 response but I'm not sure whether it's encrypted or not. It's mentioned 3 times in this doc. Could we clarify please whether an opt-out response is encrypted?**]
```
sunnywu, line 22: Optout response is encrypted according to [here ](https://github.com/IABTechLab/uid2-operator/blob/master/src/main/java/com/uid2/operator/vertx/UIDOperatorVerticle.java#L393) and [here](https://github.com/IABTechLab/uid2-operator/blob/master/src/main/java/com/uid2/operator/vertx/V2PayloadHandler.java#L132)

docs/endpoints/post-token-refresh.md
```diff
@@ -19,9 +19,8 @@ Used by: This endpoint is used mainly by publishers.
...
+- To decrypt responses, the `refresh_response_key` value returned in the [POST /token/generate](post-token-generate.md) or `POST /token/refresh` response from which the refresh token in the request is returned. {**GWH/SW_02. Please help me understand this line.**}
```
sunnywu, line 23: So basically refresh_response_key is the similar concept as "client secret" for token/generate call, that is used to encrypt/decrypt the successful response in token/refresh. So when token/generate creates a refresh token response, it has the first refresh_response_key to be used to encrypt the response of next token/refresh call and then each token/refresh response will a new refresh_response_key for the next subsequent token/refresh call response decryption and so on. 


@jon8787  please confirm if what i said makes sense about the purpose of refresh_response_key

docs/endpoints/post-token-refresh.md
```diff
@@ -31,6 +30,8 @@ Here's what you need to know about this endpoint:
 
 >NOTE: The integration environment and the production environment require different [API keys](../ref-info/glossary-uid.md#gl-api-key).
 
+{**GWH/SW_03 query. It doesn't say, anywhere, how the actual token is conveyed. Path parameter? Body? This ties in with my question #4.**}
```
sunnywu, line 33: In the `POST '{environment}/v2/token/refresh'`  on line 17 above, you can just say add the content of the refresh_token value from previously token/gen or token/refresh response as the _POST body_ and that's how you call token/refresh

docs/endpoints/post-token-refresh.md
```diff
@@ -42,9 +43,11 @@ Using either of the following parameters in a [POST /token/generate](post-token-
 
 For details and Python script examples, see [Encrypting Requests and Decrypting Responses (Python script example)](../getting-started/gs-encryption-decryption#uid2_requestpy).
 
+{**GWH/SW_04. I would dearly love to have a plain request example AS WELL AS this link to a Python script. I feel strongly that we shoudl have an example request and response in the doc for each operation. Is it possible I could get one? Fictitious of course. user@example.com would be fine.**}
```
sunnywu, line 46: Are you saying we should have a uid2_request.py example on each page of token/refresh and token/generate instead of just having 1 combined script on https://unifiedid.com/docs/getting-started/gs-encryption-decryption page?

docs/endpoints/post-token-refresh.md
```diff
@@ -95,10 +98,10 @@ An error response might look like the following:
 
 | Property | Data Type | Description |
 | :--- | :--- | :--- |
-| `advertising_token` | string | An encrypted advertising (UID2) token for the user. |
+| `advertising_token` | string | An encrypted UID2 token (advertising token) for the user. |
```
sunnywu, line 101: I think this should be along the line of:
_also known as UID2 token, it's an encrypted form of a raw UID2 representing the user._

docs/endpoints/post-token-refresh.md
```diff
@@ -108,10 +111,11 @@ The following table lists the `status` property values and their HTTP status cod
 
 | Status | HTTP Status Code | Description |
 | :--- | :--- | :--- |
-| `success` | 200 | The request was successful. The response will be encrypted. |
+| `success` | 200 | The request was successful and a new UID2 token, with associated values, is returned in the response. The response is encrypted. |
 | `optout` | 200 | The user opted out. This status is returned only for authorized requests. |
 | `client_error` | 400 | The request had missing or invalid parameters.|
-| `invalid_token` | 400 | The request had an invalid identity token specified. This status is returned only for authorized requests. |
+| `invalid_token` | 400 | The UID2 token specified in the request was invalid. This status is returned only for authorized requests. |
+| `expired_token` | 400 | The UID2 token specified in the request was an expired token. |
```
sunnywu, line 118: _The refresh token specified in the request was expired._

It's not the UID2 token but the refresh token (that contains the UID2 token) has expired

docs/endpoints/post-token-refresh.md
sunnywu: @jon8787 please give any advice if you have any

docs/endpoints/post-token-refresh.md
jon8787: Same with the Java SDK - it decrypts any 2xx response

docs/endpoints/post-token-refresh.md
jon8787: That's correct. The reason we have refresh_response_key is because we can't use the initial client secret to decrypt this. That's because decrypting /token/refresh responses happens in the user's browser, so any secret would've been visible in the JavaScript. 

"If the request is successful, with an HTTP status code of 200 and a new UID2 token or opt-out information returned, the response is encrypted." Should we split this into 2 sentences?

docs/endpoints/post-token-refresh.md
jon8787: token/generate already has an example.

For token/refresh, you could copy the refresh example from the script on that page:

> python3 uid2_request.py https://prod.uidapi.com/v2/token/refresh --refresh-token AAAAAxxJ...(truncated, total 388 chars) v2ixfQv8eaYNBpDsk5ktJ1yT4445eT47iKC66YJfb1s=

I suspect it's not very helpful since the user won't know what the refresh-token and refresh_response_key should be.

docs/endpoints/post-token-refresh.md
jon8787: Right, "UID2 token" is already encrypted

docs/endpoints/post-token-refresh.md
genwhittTTD: Done. Though, would LOVE an example of that.

docs/endpoints/post-token-refresh.md
genwhittTTD: Sunny: no just saying a plain example with no script. If that' possible. Jonathan: no prob will not change it. I was thinking maybe a cURL example or something. Something really simple.

docs/endpoints/post-token-refresh.md
genwhittTTD: Fixed. TY!

docs/endpoints/post-token-refresh.md
```diff
@@ -95,10 +99,10 @@ An error response might look like the following:
 
 | Property | Data Type | Description |
 | :--- | :--- | :--- |
-| `advertising_token` | string | An encrypted advertising (UID2) token for the user. |
+| `advertising_token` | string | The UID2 token (advertising token) for the user. |
```
sunnywu, line 102: _The UID2 token (also known as advertising token) for the user. |_
Otherwise i feel like we were saying advertising_token is advertising token ?!

sunnywu: But will leave it to you Gen :)

docs/endpoints/post-token-refresh.md
genwhittTTD: Updated. Thx.

docs/endpoints/post-token-refresh.md
genwhittTTD: I fixed this, Sunny. Used your wording basically, and added a link also.

PR #225, opened 2023-07-25
Gwh apidocs 1588 02 add sdk docs

genwhittTTD:
Updates to get the SDK doc files in place so the file sets are better in sync. Plus all broken links in the Japanese are fixed.


genwhittTTD:
This should have been a merge to Japanese branch not main. Closing 225. 227 makes these mods.

PR #232, opened 2023-08-02
Feat: Add Japanese Translations 🇯🇵 

bbuilds:
This PR enabled multi-lingual capabilities of Docusaurus and translate the website content into Japanese 

- enabled 118n on Docusaurus
- translate markdown files to Japanese
- translate React components to Japanese 
- adds a new partner set of logos


src/components/SectionHeader/index.tsx
```diff
@@ -27,7 +29,11 @@ export default function SectionHeader({
           {eyebrow}
         </p>
       )}
-      <h2 className={clsx("type-beta col col--8")}>{heading}</h2>
+      <h2 className={clsx("type-beta col col--8")}>
+        {heading}{" "}
+        {jaHeading && <span className={styles.jaHeading}>{jaHeading}</span>}
```
lionell-pack-ttd, line 34: I've noticed a few patterns like this throughout - it's fine for now, but won't really scale if we add additional translations. I think it's OK to merge like this, but we'll need a more general approach if we add any more languages.

lionell-pack-ttd (approved):

src/pages/request-access.tsx
```diff
@@ -30,6 +58,8 @@ export default function RequestDemo(): JSX.Element {
 
   const formRef = React.useRef(null);
 
+  const formId = i18n.currentLocale === "ja" ? 3688 : 2753;
```
lionell-pack-ttd, line 61: Might it be better to use the `useIsJapanese` helper here? Not a blocker to merging.

PR #278, opened 2023-08-23
state in doc that policy parameter with value 1 is required

genwhittTTD:
state in doc that policy parameter with value 1 is required + update request examples.


cody-constine-ttd (approved):
LGTM

PR #320, opened 2023-10-18
JS SDK v3 review comments

genwhittTTD:
JS SDK v3 review comments (SDK guide only, not integration guide)


docs/sdks/client-side-identity.md
```diff
@@ -117,25 +125,25 @@ Here's what you need to know about the token auto-refresh:
...
 You can provide as many callback functions as you want, and register them from anywhere. This allows you to split your code up in a way that makes sense for your site.
 :::
 
 #### Callback Function Signature
 
-Your callback function should accept two parameters: an event type and a payload. The type of the payload will depend on the event type.
+Your callback function should accept two parameters: an event type and a payload. The type of the payload depends on the event type.
 
-This is an example callback which handles the `SdkLoaded` event to call init, and the `InitCompleted` event to provide an identity if one isn't available once `init` has completed.
+The following example callback handles the `SdkLoaded` event to call `init` and then, if an identity isn't available once `init` has completed, calls the `InitCompleted` event to provide an identity.
```
lionell-pack-ttd, line 146: This re-wording isn't quite right, can we use:

> The following example callback handles the `SdkLoaded` event to call `init` and then, if an identity isn't available once `init` has completed, uses the `InitCompleted` event to provide an identity.

docs/sdks/client-side-identity.md
```diff
@@ -176,51 +184,55 @@ This is an example callback which handles the `SdkLoaded` event to call init, an
...
+- [Provide an Identity in the Call to `init`](#provide-an-identity-in-the-call-to-init)
+- [Provide an Identity by Calling `setIdentity`](#provide-an-identity-by-calling-setidentity)
 
-#### Provide an identity by setting a first-party cookie
+#### Provide an Identity by Setting a First-Party Cookie
 
-If you store a first-party cookie as described in the [storage format section](#uid2-storage-format), the SDK will load that value if it is newer than the value available in local storage. If you have set the `useCookie` init option to `true`, it will always load this value and not check local storage. You can control several other things about the cookie using [init parameters](#init-parameters).
+If you store a first-party cookie, as described in the [storage format section](#uid2-storage-format), and the value is newer than the value available in local storage, the SDK loads the value in the cookie. If you have set the `useCookie` init option to `true`, it always loads this value, and does not check local storage. You can control several other things about the cookie using [init parameters](#init-parameters).
```
lionell-pack-ttd, line 222: Could we change "the SDK loads the value in the cookie." to "the SDK loads the value from the cookie."?

"Load" can have a slightly different meaning in software which could lead someone to think this means the reverse (i.e. they might reasonably think it means that it will store the value into the cookie). I think using "from" removes any ambiguity.

docs/sdks/client-side-identity.md
```diff
@@ -232,11 +244,11 @@ All interactions with the Client-Side JavaScript SDK are done through the global
 - [abort()](#abort-void)
 - [callbacks](#callbacks) <New />
 - [setIdentity()](#setidentityidentity-uid2identity-void) <New />
-- [getIdentity()](TODO) <New />
+- [getIdentity()](#getidentity-uid2identity--null) <New />
 
 ### constructor()
 
-Constructs a UID2 object. This is not intended to be used directly: when the SDK loads, it automatically initializes an instance of the UID2 class and stores it as the global `__uid2` object. Advanced integrations may make use of this constructor directly, but care must be taken to avoid having multiple active instances of the SDK running. This is not a supported use-case.
+Constructs a UID2 object. This is not intended to be used directly: when the SDK loads, it automatically initializes an instance of the UID2 class and stores it as the global `__uid2` object. Advanced integrations could make use of this constructor directly, but be careful to avoid having multiple active instances of the SDK running. This is not a supported use case.
```
lionell-pack-ttd, line 251: I'd like to keep using the words `may` and `must` here because they have special meaning in software engineering (see https://www.ietf.org/rfc/rfc2119.txt for details - while that document is specific to the meaning in IETF documents, the IETF is the tech industry's main standards body so the meaning is generally considered universal in tech industry documents).

Perhaps:

> Constructs a UID2 object. This is not intended to be used directly: when the SDK loads, it automatically initializes an instance of the UID2 class and stores it as the global `__uid2` object. Advanced integrations may make use of this constructor directly, but must take care to avoid having multiple active instances of the SDK running. This is not a supported use case.

docs/sdks/client-side-identity.md
```diff
@@ -248,11 +260,11 @@ Initializes the SDK and establishes user identity for targeted advertising.
...
-- If the `identity` property in the `init()` call is falsy, the SDK will attempt to load the identity from local storage or the cookie.
-  - Once `init()` is complete, all callbacks will receive the `InitCompleted` event. If the `identity` property on the payload of this event is null, no identity could be loaded and you should [provide a valid identity](#provide-identity). This is the recommended way to provide an identity if your server-side integration does not ensure a current identity is always available, and you need to request it from the server only when necessary.
-  - If you are using a [first-party cookie](#uid2-cookie-format) to store the passed UID2 information for the session, a call to `init()` made by a page on a different domain might not be able to access the cookie. You can adjust the settings used for the cookie with the `cookieDomain` and `cookiePath` options.
+- If the `identity` property in the `init()` call is false, the SDK attempts to load the identity from local storage or the cookie.
```
lionell-pack-ttd, line 265: This must be `falsy` and not `false` - it's a specific JavaScript/TypeScript language term which means something specific.

docs/sdks/client-side-identity.md
```diff
@@ -431,28 +447,28 @@ Returns the current stored identity, if available.
...
+### UID2 Cookie Properties
 
-If cookie storage is being used, the cookie will use the properties in this table.
+If cookie storage is being used, the cookie uses the properties in the following table.
 
 | Properties | Default Value | Comments |
 | :--- | :--- | :--- |
 | `Name` | `__uid_2` | N/A |
-| `Expiry` | N/A | The value is the refresh token expiration timestamp as specified in the [POST /token/generate](../endpoints/post-token-generate.md) or [POST /token/refresh](../endpoints/post-token-refresh.md) response. |
+| `Expiry` | N/A | The value is the refresh token expiration timestamp as specified in the<br/>[POST /token/generate](../endpoints/post-token-generate.md) or [POST /token/refresh](../endpoints/post-token-refresh.md) response. |
```
lionell-pack-ttd, line 465: Is the `<br/>` intentional here? It can put the line break in an unfortunate place at certain browser widths, e.g.:
image

docs/sdks/client-side-identity.md
lionell-pack-ttd: See https://developer.mozilla.org/en-US/docs/Glossary/Falsy for details if you're interested 🙂

docs/sdks/client-side-identity.md
genwhittTTD: Done. Thx!

docs/sdks/client-side-identity.md
genwhittTTD: Done thx! And good to know. Generally I try not to use "may" because it's ambiguous -- it can express possibility (it may rain today -- I use "might") or permission (as an administrator you may add a user -- I use "can"). I think I picked the wrong one here. But anyway it now says exactly what you suggested. Thx much.

docs/sdks/client-side-identity.md
genwhittTTD: Thx! Removed it. Yes it was intentional, and it looked good to me. My bad.

lionell-pack-ttd (approved):

PR #329, opened 2023-10-26
copy review on JS SDK integration guide w/queries marked

genwhittTTD:
copy review on JS SDK integration guide w/queries marked (4)

### DIFF
--- a/docs/guides/publisher-client-side.md
+++ b/docs/guides/publisher-client-side.md
@@ -37,10 +37,8 @@ For a workflow diagram, see [Integration Steps](#integration-steps). See also [F
 
 To facilitate the process of establishing client identity using UID2 and retrieving advertising tokens, the web integration steps provided in this guide rely on the UID2 SDK for JavaScript. Here's an [example application](https://example-jssdk-integ.uidapi.com/) that illustrates the integration steps described in this guide and the usage of the SDK (currently only for email addresses). For the application documentation, see [UID2 SDK Integration Example](https://github.com/IABTechLab/uid2-examples/blob/main/publisher/standard/README.md).
 
-(**GWH_LP01 any way we could go a little gentler than Danger? It looks scary. Maybe just a Note? But, your call.**)
-
-:::warning
-The UID2 SDK for JavaScript currently stores tokens in first-party cookies, but implementation details such as this might change in the future. To avoid potential issues, be sure to rely on the [Client-Side JavaScript SDK](../sdks/client-side-identity.md#api-reference) for your identity management.
+:::tip
+The first-party cookie and local storage implementation details might change in the future. To avoid potential issues, be sure to rely on the functionality documented in the [UID2 SDK for JavaScript API Reference](../sdks/client-side-identity.md#api-reference) for your identity management.
 :::
 
 For integration scenarios for publishers that do not use the UID2 SDK for JavaScript, see [Publisher Integration Guide, Server-Only](custom-publisher-integration.md). 
@@ -72,9 +70,7 @@ After authentication in step 1-c, which forces the user to accept the rules of e
 | 1-d | [POST /token/generate](../endpoints/post-token-generate.md) | After the user authenticates and authorizes the creation of a UID2, use the [POST /token/generate](../endpoints/post-token-generate.md) endpoint to generate a UID2 token using the email address or phone number provided by the user. Make sure it is normalized. |
 | 1-e | [POST /token/generate](../endpoints/post-token-generate.md) | The endpoint returns a UID2 token generated from the user's email address, phone number, or the respective hash. |
 | 1-f | UID2 SDK for JavaScript | The SDK sends the returned UID2 token from step 1-e to the SDK in the `identity` property of its [init() function](../sdks/client-side-identity.md#initopts-object-void). |
-| 1-g | UID2 SDK for JavaScript | Provide a callback function to the SDK, which will receive identity updates and use them to initiate targeted advertising. |
-
-(**GWH_LP02 Hoping to clarify 1-g above. The SDK is listed as the actor but the statement says "Provide a callback function to the SDK" -- could we word this better? Maybe something like, set up the SDK's callback function to receive...?**)
+| 1-g | UID2 SDK for JavaScript | Provide the SDK a callback function that will receive identity updates from the SDK and use them to initiate targeted advertising. |
 
 <Tabs>
 <TabItem value='js' label='JavaScript'>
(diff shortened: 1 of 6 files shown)


docs/guides/publisher-client-side.md
```diff
@@ -29,12 +31,16 @@ This guide is intended for publishers with web assets who want to generate ident
...
+This guide outlines the basic steps that you need to consider if you are building an integration without using an SDK. For example, you need to decide how to implement user login and logout, how to manage UID2 identity information and use it for targeted advertising, and how to refresh tokens, deal with missing identities, and handle user opt-outs.
+
+For a workflow diagram, see [Integration Steps](#integration-steps). See also [FAQs](#faqs).
 
 To facilitate the process of establishing client identity using UID2 and retrieving advertising tokens, the web integration steps provided in this guide rely on the UID2 SDK for JavaScript. Here's an [example application](https://example-jssdk-integ.uidapi.com/) that illustrates the integration steps described in this guide and the usage of the SDK (currently only for email addresses). For the application documentation, see [UID2 SDK Integration Example](https://github.com/IABTechLab/uid2-examples/blob/main/publisher/standard/README.md).
 
+(**GWH_LP01 any way we could go a little gentler than Danger? It looks scary. Maybe just a Note? But, your call.**)
```
lionell-pack-ttd, line 40: Yeah - I think something less strong would be fine, maybe a tip. The paragraph doesn't even quite make sense - it tells you that the SDK stores tokens in first-party cookies, so make sure you use the SDK for identity management (but you already are, it's the thing storing the cookie!)

Suggested change:

> The first-party cookie and local storage implementation details might change in the future. To avoid potential issues, be sure to rely on the functionality documented in the [UID2 SDK for JavaScript API Reference](../sdks/client-side-identity.md#api-reference) for your identity management.

docs/guides/publisher-client-side.md
```diff
@@ -45,28 +51,30 @@ If you are using Google Ad Manager and want to use the secure signals feature, f
...
+| 1-d | [POST /token/generate](../endpoints/post-token-generate.md) | After the user authenticates and authorizes the creation of a UID2, use the [POST /token/generate](../endpoints/post-token-generate.md) endpoint to generate a UID2 token using the email address or phone number provided by the user. Make sure it is normalized. |
+| 1-e | [POST /token/generate](../endpoints/post-token-generate.md) | The endpoint returns a UID2 token generated from the user's email address, phone number, or the respective hash. |
+| 1-f | UID2 SDK for JavaScript | The SDK sends the returned UID2 token from step 1-e to the SDK in the `identity` property of its [init() function](../sdks/client-side-identity.md#initopts-object-void). |
+| 1-g | UID2 SDK for JavaScript | Provide a callback function to the SDK, which will receive identity updates and use them to initiate targeted advertising. |
+
+(**GWH_LP02 Hoping to clarify 1-g above. The SDK is listed as the actor but the statement says "Provide a callback function to the SDK" -- could we word this better? Maybe something like, set up the SDK's callback function to receive...?**)
```
lionell-pack-ttd, line 77: Is this clearer?

> Provide the SDK a callback function which will receive identity updates and use them to initiate targeted advertising.

The intent is that they need to write the callback function (which receives identity updates and initiates targeted advertising - that's their code). The SDK provides the identity updates, but it's up to the publisher to do something with them.

docs/guides/publisher-client-side.md
```diff
@@ -153,11 +161,19 @@ Depending on the structure of your code, it might be convenient to combine the c
...
 ### Bid Using UID2 Tokens
 
-Based on the status and availability of a valid identity, the SDK sets up the background token auto-refresh, stores identity information in [local storage or a first-party cookie](../sdks/client-side-identity.md#uid2-storage-format), and uses it to initiate requests for targeted advertising.
+Based on the status and availability of a valid identity, the SDK does the following:
+
+1. Sets up the background token auto-refresh.
+1. Stores identity information in [local storage or a first-party cookie](../sdks/client-side-identity.md#uid2-storage-format).
+1. Uses the identity information to initiate requests for targeted advertising.
+
+(**GWH_LP03 question. Not sure about the relationship between the steps above and the table below. And the diagram 2-a which says "the publisher calls the SSP for ads using the UID2 token". Any way you could help me understand/clarify? They don't seem entirely in sync.**)
```
lionell-pack-ttd, line 170: It looks like the diagram needs to be updated. I think maybe once we've finished any other changes, we should re-do it.

docs/guides/publisher-client-side.md
```diff
@@ -171,18 +187,20 @@ Based on the status and availability of a valid identity, the SDK sets up the ba
 You need to consider how you pass the returned advertising token to SSPs.
 :::
 
+(**GWH_LP04 is there any guidance at all that we can offer re the above? Maybe an example or something?**)
```
lionell-pack-ttd, line 190: The difficulty here is that our main advice is to use Prebid.js (in which case they shouldn't be following this guide) or Google Ad Manager Secure Signals (in which case they shouldn't be following this guide).
The only publishers using this guide should be those who *don't* want to follow our advice about how to pass the token to SSPs - probably because they already have a different system they use.

Perhaps some text to that effect might be helpful here?

docs/guides/publisher-client-side.md
genwhittTTD: Awesome thx fixed.

docs/guides/publisher-client-side.md
genwhittTTD: TY! Here's what I've put, LMK what you think. We can tweak it any way you want or go back to the original.
"Provide the SDK a callback function that will receive identity updates from the SDK and use them to initiate targeted advertising."

docs/guides/publisher-client-side.md
genwhittTTD: Sounds good. No change here for now then! Thx.

docs/guides/publisher-client-side.md
genwhittTTD: Got it. Thx. Here's what I put. Any input welcome.
You need to consider how you pass the returned advertising token to SSPs. With some other approaches to client-side UID2 implementation, such as using Prebid.js (see Prebid Integration Guide or Google Ad Manager Secure Signals (see Google Ad Manager Secure Signals Integration Guide), the implementation includes functions that manage passing the returned advertising token. If you're using the UID2 SDK for JavaScript you'll need to manage this yourself.

lionell-pack-ttd:
Looks good to me :) merging as requested.

PR #356, opened 2023-11-09
feat: update deeply nested nav css

bbuilds:
- reduce margin on sub list parents
- reduce padding on 3rd level nested items


genwhittTTD (approved):
TYSM!

PR #358, opened 2023-11-13
UID2-2231 created javascript express integration guide and fixed a missing word  in the prebid guide

sunnywu:
1. Calling the new guide JavaScript Express Integration Guide
2. Renaming the old UID2 SDK for JavaScript Integration Guide to JavaScript Standard Integration Guide. We [used ](https://github.com/IABTechLab/uid2docs/pull/358/files#diff-107e9a50e50335a1f5bbe6a361d3921b3bd45e140eb5c90991e71e163e117ba9R170)the term Standard Integration in python sdk (for better or worse). I don't wanna call it Advanced Guide like [prebid.js' ](http://prebid.as/)as this is probably more widely used and don't wanna deter people away thinking it's too complicated.
3. I don't think the term UID2 SDK for Javascript adds any value to the guide name (as we clearly have 2 guides now) and it is important to differentiate between the Express and Standard (hence renaming to JavaScript XXXX Integration Guides).
4. I also removed the term Client-Side as people will realize it's client side as it's JavaScript and after reading the first paragraph in the guide.
5. I have changed the high level descriptions of the Standard Integration Guide in various overview pages to mention the difference is server side token generation in order to clearly differentiate with the new Express Guide.


docs/guides/integration-javascript.md
```diff
@@ -0,0 +1,293 @@
...
+
+With the setup above, when the UID2 SDK generates a UID2 token, the advertising token (TODO add link to glossary) will be ready to retrieve by calling the ```getAdvertisingToken```. The advertising token is the token publishers will pass to the bidstream. Before calling this function, be sure to call init() and wait until your callback handler has received an InitCompleted event.
+
+```js
+<script>
+  let advertisingToken = __uid2.getAdvertisingToken();
+</script>
+```
+
+See more info and typescript example on [getAdvertisingToken API reference](../sdks/client-side-identity#getadvertisingtoken-string)
```
alex-yau-ttd, line 236: typescript -> TypeScript

docs/guides/integration-javascript.md
```diff
@@ -0,0 +1,293 @@
...
+
+There is another call ```getAdvertisingTokenAsync``` which can return a ```Promise``` string for the latest advertising token.   
+
+See more info and typescript example on [getAdvertisingTokenAsync API reference](../sdks/client-side-identity#getadvertisingtokenasync-promise)
```
alex-yau-ttd, line 240: typescript -> TypeScript?

docs/guides/integration-javascript.md
```diff
@@ -0,0 +1,293 @@
+---
+title: Javascript Express Integration
```
alex-yau-ttd, line 2: Javascript -> JavaScript. Same comment for all other usages of "Javascript"

docs/guides/integration-javascript.md
```diff
@@ -0,0 +1,293 @@
...
+
+When the UID2 SDK is configured, it checks for an existing UID2 token in the user's browser. If there is an existing token that was generated from the same DII, and the token is still valid or can be refreshed, the SDK uses or refreshes the existing token instead of generating a new token.
+
+If there is no existing token, or the token has expired and cannot be refreshed, the UID2 SDK cannot generate a new token without DII.
+
+As a result, the recommended approach is to call the ```setIdentityFromXXXX``` method with the user's DII on each page load.
+
+In some cases, the user's DII is not available on page load, and getting the DII has some associated cost. For example, an API call might be required to fetch the DII, or the user has to be prompted to provide the DII information.
+
+You can potentially avoid that cost by checking for an existing token that you can use or refresh - simply calling
```
sunnywu, line 262: is this the correct suggestion @lionell-pack-ttd ?

docs/guides/integration-javascript.md
sunnywu: fixed

docs/guides/integration-javascript.md
sunnywu: fixed

docs/guides/integration-javascript.md
sunnywu: fixed

docs/guides/integration-javascript.md
```diff
@@ -0,0 +1,293 @@
...
+Support for client-side token generation is available in version 3.2 and above of the SDK. 
+
+The URL for the SDK is https://cdn.prod.uidapi.com/uid2-sdk-3.2.0.js . In the following code examples, we will use the placeholder ```{{ UID2_JS_SDK_URL }}``` to refer to this URL.
+
+If you want to use a debug build of the SDK, use the following URL instead: https://cdn.integ.uidapi.com/uid2-sdk-3.2.0.js . 
+
+## Sample Implementation Website
+
+For an example application, see this example:
+- Running site: [UID2 SDK Integration Example](https://cstg-integ.uidapi.com/)
```
lionell-pack-ttd, line 41: A link to the code would be good as well, especially as the example site doesn't have a link to it.

docs/guides/integration-javascript.md
```diff
@@ -0,0 +1,293 @@
...
+When account setup is complete, you'll receive a **public key** and **subscription ID**. These values are unique to you, and you'll use them to configure the UID2 module.
+
+:::tip
+Only root-level domains are required for account setup. For example, if you're going to use UID2 SDK for JavaScript on example.com, shop.example.com, and example.org, you only need to provide the domain names example.com and example.org.
+:::
+
+
+## Add UID2 SDK For JavaScript to Your Site
+
+The following code illustrates the different SDK events that we will respond to.
```
lionell-pack-ttd, line 56: They're events that the SDK can generate - it's their code that can choose to respond to them.

Perhaps "The following code illustrates the different events that the SDK can trigger."?

docs/guides/integration-javascript.md
```diff
@@ -0,0 +1,293 @@
...
+      break;
+ 
+    case "InitCompleted":
+      // The InitCompleted event occurs just once.
+      //
+      // If there was a valid UID2 token, it will be in payload.identity.
+      break;
+ 
+    case "IdentityUpdated":
+      // The IdentityUpdated event will happen when a UID2 token was generated or refreshed.
```
lionell-pack-ttd, line 80: I think it's worth adding that payload.identity will contain a valid UID2 token if one is available for this event too.

docs/guides/integration-javascript.md
```diff
@@ -0,0 +1,293 @@
...
+Only root-level domains are required for account setup. For example, if you're going to use UID2 SDK for JavaScript on example.com, shop.example.com, and example.org, you only need to provide the domain names example.com and example.org.
+:::
+
+
+## Add UID2 SDK For JavaScript to Your Site
+
+The following code illustrates the different SDK events that we will respond to.
+
+```js
+<script defer src="{{ UID2_JS_SDK_URL }}"></script>
```
lionell-pack-ttd, line 59: If we're going to recommend `defer`, I'd much prefer to use `async` instead - which will result in the script being loaded and run sooner.

docs/guides/integration-javascript.md
```diff
@@ -0,0 +1,293 @@
...
+## Configure the UID2 SDK For JavaScript
+
+UID2 will provide the publisher with the following values required to use the client-side token generation feature:
+
+* A subscription ID, and
+* A public key.
+
+There will be one set of these values for the publisher's testing environment(s), and a separate set for the publisher's production environment.
+
+To configure the SDK, call `setIdentityFromXXXX` with an object containing the **public key** and **subscription ID** that you received during account setup, as well as the user's hashed or unhashed [DII](../ref-info/glossary-uid.md#gl-dii) (email address or phone number).
```
lionell-pack-ttd, line 130: I think it would be better to list the options instead of saying `setIdentityFromXXXX`.

lionell-pack-ttd: Or maybe not list all four, but I don't like `setIdentityFromXXXX` much.

docs/guides/integration-javascript.md
```diff
@@ -0,0 +1,293 @@
...
+);
+```
+
+**The publisher is responsible for normalizing and hashing the phone number**. For details, see [Normalization and Encoding](../getting-started/gs-normalization-encoding.md).
+
+The UID2 SDK encrypts the hash before sending it to the UID2 service.
+
+## Token Storage and Refresh
+
+The generated UID2 token will be stored in local storage, under the key UID2-sdk-identity. The SDK will refresh the UID2 token periodically.
```
lionell-pack-ttd, line 224: The whole identity (the object returned from the token generate/CSTG call) is stored under UID2-sdk-identity.

I think it's worth adding a warning that the format used for this storage may change without notice, and they should not write code which reads or updates this local storage key.

docs/guides/integration-javascript.md
```diff
@@ -0,0 +1,293 @@
...
+
+## Retrieving the UID2 Advertising Token 
+
+With the setup above, when the UID2 SDK generates a UID2 token, the advertising token (TODO add link to glossary) will be ready to retrieve by calling the ```getAdvertisingToken```. The advertising token is the token publishers will pass to the bidstream. Before calling this function, be sure to call init() and wait until your callback handler has received an InitCompleted event.
```
lionell-pack-ttd, line 228: It's probably better to use the events mentioned above to receive the token - if they rely on `getAdvertisingToken` they won't receive changes to it and could end up with an old token even though a refreshed one becomes available straight after.

docs/guides/integration-javascript.md
lionell-pack-ttd: I think we should be providing examples based on the events - but we don't have a good event for this case yet. If `InitCompleted` contains no identity, there might still be an active request trying to load one.

`getAdvertisingToken` provides no useful information at all about this. The only viable option right now is `getAdvertisingTokenAsync`.

docs/guides/integration-javascript.md
sunnywu: changed to mention waiting for the IdentityUpdated event

docs/guides/integration-javascript.md
sunnywu: i have given it a shot to provide examples and explanation. please have a read again and see if it makes sense.

docs/intro.md
```diff
@@ -103,12 +103,12 @@ The following table lists the key participants and their roles in the UID2 [work
...
+| [Workflow for Publishers](overviews/overview-publishers.md#workflow-for-publishers) | Organizations that propagate UID2 tokens to the bid stream via SSPs.<br/> NOTE: Publishers can choose to leverage the UID2 SDK for JavaScript (see [JavaScript Express Integration Guide](publisher-client-side.md) or [UID2 SDK for JavaScript Reference Guide](sdks/client-side-identity.md)) or complete their own server-only integration without using an SDK. | [JavaScript Express Integration Guide](publisher-client-side.md)<br/>[JavaScript Standard Integration Guide](guides/integration-javascript-standard.md)<br/>[Publisher Integration Guide, Server-Only](guides/custom-publisher-integration.md) |
```
sunnywu, line 110: @genwhittTTD why doesn't this Workflow For Publishers section include Prebid.js guides? Is it intentional?

docs/guides/integration-javascript.md
sunnywu: changed to make async call now

docs/intro.md
genwhittTTD: Hi @sunnywu good spot. That was an omission. That home page has had very few updates. We should probably revisit the whole thing. Anyway I added Prebid. Took the links out of col 2 since they are in col 3. Though, I'm thinking that col 3 should probably link to the resources section of the publisher overview page. Trying to stick to the job at hand though.

docs/guides/integration-javascript.md
lionell-pack-ttd: The explanation and example looks good - but see Gen's comment, it does look like just one snippet but the explanation says "snippets" plural.

docs/guides/publisher-client-side.md
```diff
@@ -55,14 +60,16 @@ Only root-level domains are required for account setup. For example, if you're g
 
 ## Add UID2 SDK For JavaScript to Your Site
 
+{**GWH_SW_02: The title indicates that this is a step, so I'd expect some instruction here. Did we miss anything? Like, add this code to your site? (where). Or any other sort of guidance, like if this add this bit of code, if some other scenario add a different bit?**}
```
sunnywu, line 63: _{**GWH_SW_02: The title indicates that this is a step, so I'd expect some instruction here. Did we miss anything? Like, add this code to your site? (where). Or any other sort of guidance, like if this add this bit of code, if some other scenario add a different bit?**}_

rephrased to say the code below is something pubs need to copy but a more thorough example will be provided further below.

docs/guides/publisher-client-side.md
```diff
@@ -231,29 +239,31 @@ The UID2 SDK encrypts the hash before sending it to the UID2 service.
...
+You can potentially avoid that cost by checking for an existing token that you can use or refresh. To do this, call
+[__uid2.isLoginRequired](../sdks/client-side-identity#isloginrequired-boolean) which returns a Boolean value. If it returns `true`, this means that the UID2 SDK cannot create a new advertising token with the existing resource and DII is required to generate a brand new UID2 token.
+
+The following code snippets demonstrate how you might integrate with the UID2 SDK for JavaScript for the two scenarios above&#8212;starting with no token as well as reusing/refreshing any existing UID2 token if found. 
 
-The following code snippets demonstrate how you might integrate with UID2 SDK for JavaScript for the two scenarios above - starts with no token as well as reusing/refresh any existing UID2 token if found. 
+{**GWH_SW_03 is there more than one snippet? Looks like one to me. Could we break it up and have a text intro for each piece? Maybe not a good idea. Just checking**}
```
sunnywu, line 259: _{**GWH_SW_03 is there more than one snippet? Looks like one to me. Could we break it up and have a text intro for each piece? Maybe not a good idea. Just checking**}_

yeh only 1 snippet. i think a single block is fine so readers can have a holistic view on it

docs/guides/publisher-client-side.md
```diff
@@ -318,20 +328,20 @@ window.__uid2.callbacks.push(async (eventType, payload) => {
...
+If there was a problem generating the token, find the request in the **Network** tab. You can find the request by filtering for the string `client-generate`. Information about why the request failed should be available in the response.
 
 ![Publisher Workflow](images/NetworkTraffic.png)
 
 
-## Example codes: Hashing and Base64-encoding
+## Example codes: Hashing and Base64 Encoding
 
-The following codes are examples on how to generate email and phone hashes in JavaScript.
+The following codes are examples of how to generate email and phone hashes in JavaScript. {**GWH_SW_04 should I put this into the Normalization and Encoding article?**}
```
sunnywu, line 344: _{**GWH_SW_04 should I put this into the Normalization and Encoding article?**}_

yes please and maybe add an explicit link to these js example codes in the "Configure the UID2 SDK for JavaScript" section above?

docs/intro.md
```diff
@@ -107,8 +107,8 @@ The following table lists four key workflows in the UID2 framework with links to
...
+| [Workflow for Publishers](overviews/overview-publishers.md#workflow-for-publishers) | Organizations that propagate UID2 tokens to the bid stream via SSPs.<br/> NOTE: Publishers can choose to integrate using Prebid, leverage the UID2 SDK for JavaScript, or complete their own server-only integration without using an SDK. | [Prebid.js Express Integration Guide](guides/integration-prebid.md)<br/>[JavaScript Express Integration Guide](guides/publisher-client-side.md)<br/>[JavaScript Standard Integration Guide](guides/integration-javascript-standard.md)<br/>[Publisher Integration Guide, Server-Only](guides/custom-publisher-integration.md) |
```
sunnywu, line 110: @genwhittTTD the Prebid.js advanced guide is missing here?

docs/guides/integration-javascript.md
sunnywu: thanks i fixed it

docs/guides/publisher-client-side.md
```diff
@@ -22,18 +22,23 @@ UID2 provides a UID2 SDK for JavaScript (TODO add links!!!) with the following f
 You'll need to complete the following steps:
 
 1. [Complete UID2 account setup](#complete-uid2-account-setup)
-2. [Add UID2 SDK For JavaScript to Your Site](#add-js-to-your-site)
-3. [Configure the UID2 SDK For JavaScript](#configure-the-uid2-sdk-for-javascript)
-4. [Retrieving the UID2 Advertising Token](#retrieving-the-uid2-advertising-token)
-
+2. [Add UID2 SDK For JavaScript to your site](#add-js-to-your-site)
+3. [Configure the UID2 SDK for JavaScript](#configure-the-uid2-sdk-for-javascript)
+4. [Check that the token was successfully generated](#check-that-the-token-was-successfully-generated) {**GWH_SW_01: there is no heading/link for the step as it was. Wasn't sure what it should link to. I've done what I thought might be correct.**}
```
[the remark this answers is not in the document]
genwhittTTD: Fixed, thx.

docs/guides/publisher-client-side.md
genwhittTTD: @sunnywu perfect, thx!

docs/guides/publisher-client-side.md
genwhittTTD: Thx. Changed the intro copy from "The following code snippets demonstrate..." to The following code snippet demonstrates...". Removed my query.

docs/guides/publisher-client-side.md
genwhittTTD: @sunnywu Done (both of the above) thx.

docs/intro.md
genwhittTTD: @sunnywu good catch thx. But that doc list is so dense. Instead in that column I replaces with a link to the applicable section in the integration guides summary page. Simpler. I did it even for DSP which only has one guide... for extensibility, we will have more!

docs/guides/publisher-client-side.md
sunnywu: @genwhittTTD i have reworded the sentences that the code snippet here is just for overview understand and explicitly tell the reader to copy the Example codes in **_Example Integration Code and When to Pass DII to the UID2 SDK section_**. Think this is good enough for now but feel free to edit my sentences further.

PR #466, opened 2024-02-22
Java SDK Usage for Advertisers and Data Providers

jon8787:


docs/sdks/uid2-sdk-ref-java.md
```diff
@@ -286,6 +286,39 @@ If you're using server-only integration (see [Publisher Integration Guide, Serve
...
+   ```
+
+
+2. Call a function that takes users' email addresses or phone numbers as input and generates an IdentityMapResponse object. The following example uses email addresses:
+   ```java
+   IdentityMapResponse identityMapResponse = identityMapClient.generateIdentityMap(IdentityMapInput.fromEmails(Arrays.asList("email1@example.com", "email2@example.com")));
+   ```
+
+
+>Note: Inputs will be hashed by the SDK before being sent. This ensures that raw email addresses and phone numbers do not leave your server.
```
sunnywu, line 302: is it better to use the ":::important .... :::" tag to make this point stand out?

jon8787: I like the idea of making it stand out, but it's not important in the sense that the user needs to do anything. More that it's a good feature for them to know about

sunnywu: ok, will see if Gen has any suggestions

sunnywu (approved):

docs/sdks/uid2-sdk-ref-java.md
genwhittTTD: My comments: First FYI I have to do a pass on the entire docs set to go to Docusaurus admonitions but waiting until EUID is done. Until then, the fewer the better because it's another change we have to make when porting content to GitHub.
On this one -- Warning is the Docu next level up from Note... and is probably not applicable here. So I think Note is good here.

Couple of other comments:

a) line 296
Call a function that takes users' email addresses or phone numbers as input > Call a function that takes email addresses or phone numbers as input
Reasoning: it's not needed, and the s' is no longer mainstream language so I leave it out if it isn't needed.

b) line 302:
Inputs will be hashed by the SDK before being sent > The SDK hashes input values before sending them
Inputs are hashed by the SDK before being sent

c) Also could we change two empty lines to one? line 295, 301.

docs/sdks/uid2-sdk-ref-java.md
```diff
@@ -63,7 +63,7 @@ The binary is published on the Maven repository:
 
 - [https://central.sonatype.com/artifact/com.uid2/uid2-client](https://central.sonatype.com/artifact/com.uid2/uid2-client)
 
-## Initialization
+## Usage for DSPs
```
genwhittTTD, line 66: Not sure about this change. Is all the content, right up to Usage for Publishers, only applicable to DSPs? Interface, Response Content, Response Statuses? At least some of that seems general. If it's all only applicable to DSPs,  not publishers, then this is fine.

genwhittTTD (commented):
Left just a couple of comments

docs/sdks/uid2-sdk-ref-java.md
sunnywu: FOr publisher/sharer/advertiser/data provider, now they have their instruction section so this bit is really only for DSP now hence we propose the section name change or otherwise DSP doesn't know what to read

docs/sdks/uid2-sdk-ref-java.md
jon8787: thanks, made those changes

docs/sdks/uid2-sdk-ref-java.md
genwhittTTD: @sunnywu thx.

PR #472, opened 2024-02-27
Update docs to match changed naming.

lionell-pack-ttd:


genwhittTTD (approved):
Awesome. Thanks for picking up all the Japanese fixes also! :-)

PR #495, opened 2024-03-26
update Standard Integration wording in server-side SDK docs

genwhittTTD:
update Standard Integration wording in server-side SDK docs


jon8787 (approved):
approved - but are we defining these terms somewhere?

jon8787 (approved):

PR #506, opened 2024-04-02
Fix wording - UID2 token should be raw UID2

jaredfrees-pg:
Let me know if I am misunderstanding the wording here.


atarassov-ttd:
lgtm, thank you!

PR #514, opened 2024-04-08
feat: :mag: stop indexing of URL params

bbuilds:
Looks like some weird URLs related to careers job postings are being generated and creating "new" URLs that google is crawling and linking to UID2 which google search console is flagging. While we do have canonical URLs in place which already solves this issue,  we can also disallow the pages from being indexed. 

`Disallow: /*?`

This will effectively prevent Google from crawling any parameterized URL in the future which should lead to fewer of these issues popping up which has been added to the robots.txt

### DIFF
--- a/static/robots.txt
+++ b/static/robots.txt
@@ -1,4 +1,7 @@
 User-agent: *
 Allow: /
 
+# Disallowing indexing or URLs with query parameters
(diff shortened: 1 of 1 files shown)


genwhittTTD (commented):
LGTM

PR #518, opened 2024-04-10
UID2-3105 Specify the domain name required for Bidstream DecryptTokenIntoRawUid method

sunnywu:


docs/sdks/uid2-sdk-ref-csharp-dotnet.md
```diff
@@ -140,7 +140,7 @@ var client = new BidstreamClient(UID2_BASE_URL, UID2_API_KEY, UID2_SECRET_KEY);
 client.Refresh();
 ```
 
-3. Decrypt a token into a raw UID2. Pass the domain name of the site where the bid originated from:
+3. Decrypt a token into a raw UID2. Pass the domain name of the site where the bid originated from. The domain name must be all lower cased without spaces and without subdomain, e.g. for ```Subdomain.DOMAIN.com``` , ```domain.com``` should be passed instead):
```
genwhittTTD, line 143: 3. Decrypt a token into a raw UID2. Pass the token, and the domain name of the site where the bid originated from. The domain name must be all lower case, without spaces and without subdomain. For example, for `Subdomain.DOMAIN.com` , pass domain.com` instead:

(minor wording mods plus inline code tagging is single ticks, whole line is triple ticks)

genwhittTTD (approved):
minor copy mods plus MD mods.

docs/sdks/uid2-sdk-ref-csharp-dotnet.md
sunnywu: updated thx

PR #529, opened 2024-04-17
Upgraded versions for iOS SDk, GMA/IMA plugins after privacy manifest file update

sunnywu:


genwhittTTD (approved):
All good plus couple nice editorial catches there, thanks much!

PR #571, opened 2024-05-20
UID2-3407 Add details about logs in aws private operator

cYKatherine:


docs/guides/operator-guide-aws-marketplace.md
```diff
@@ -252,6 +252,69 @@ Here's what you need to know about upgrading:
...
 
+## Logs
+### Where to read logs
+To access the logs, the participants need to ssh into the EC2 instance and access the logs.
+
+The logs are located at `/var/logs/` . The logs are in the format of `operator.log-<timestamp rotated>` . 
+
+### Default log settings
+#### logrotate config
+- After deploying the operator instance, the default log rotation settings will be applied:
```
genwhittTTD, line 263: We use present tense instead of future if possible. Suggest: 
When the operator instance has been deployed, the default log rotation settings are applied:
(even though this has passive in the first part, I think it's the clearest.)

docs/guides/operator-guide-aws-marketplace.md
```diff
@@ -252,6 +252,69 @@ Here's what you need to know about upgrading:
...
+    - 30 rotations of logs will be kept
```
genwhittTTD, line 264: Can we say this some other way? But I don't totally understand it. And, ending punctuation, and present tense. Also ideal not to start a sentence with a number. A couple of guesses:
- Log rotations are archived for 30 instances.
- Back log rotations are stored for 30 instances before deletion.

docs/guides/operator-guide-aws-marketplace.md
```diff
@@ -252,6 +252,69 @@ Here's what you need to know about upgrading:
...
+    - Normally if no abnormal logs, it should be equivalent to 30 days' worth of logs as it will be rotated daily
```
genwhittTTD, line 265: Suggest:
- If there are no abnormal log entries, logs are rotated daily, so log history is equivalent to 30 days of data.

Also this made me wonder what would happen if it's abnormal. That's answered in the next point but, actually, it's just if it's abnormally large. So perhaps we could say that:
- If log entries are not abnormally large, (etc)

docs/guides/operator-guide-aws-marketplace.md
```diff
@@ -252,6 +252,69 @@ Here's what you need to know about upgrading:
...
+    - If abnormality is encountered and the log size increased abnormally, the log will be rotated once it reaches 30MB
```
genwhittTTD, line 266: (also space between number and MB) Suggest:

- If log entries are very large, and the log size reaches 30 MB within a 24-hour period, the log is rotated at that point.

docs/guides/operator-guide-aws-marketplace.md
```diff
@@ -252,6 +252,69 @@ Here's what you need to know about upgrading:
...
+- The default logrotate settings is below (defined in /etc/logrotate.d/uid2operator.conf ):
```
genwhittTTD, line 267: Suggest:

- The following are the default logrotate settings, defined in `/etc/logrotate.d/uid2operator.conf`:

docs/guides/operator-guide-aws-marketplace.md
```diff
@@ -252,6 +252,69 @@ Here's what you need to know about upgrading:
...
+        maxsize 30M
+        dateext dateformat -%Y-%m-%d-%s
+        notifempty
+        sharedscripts
+        postrotate
+                /usr/sbin/syslog-ng-ctl reload
+        endscript
+}
+```
+- The explanation of the above config can be found https://linux.die.net/man/8/logrotate or running logrotate man in the linux environment
```
genwhittTTD, line 282: Linux, initial cap.
Suggest:
For a detailed explaination of this config, see [logrotate(8) - Linux man page](https://linux.die.net/man/8/logrotate) or run `logrotate` man in the Linux environment.

docs/guides/operator-guide-aws-marketplace.md
```diff
@@ -252,6 +252,69 @@ Here's what you need to know about upgrading:
...
+
+#### cronjob config
+- The logrotate will have the below script generated in /etc/cron.daily by default:
```
genwhittTTD, line 285: Suggest:
- The logrotate generates the following script in /etc/cron.daily by default:

docs/guides/operator-guide-aws-marketplace.md
```diff
@@ -252,6 +252,69 @@ Here's what you need to know about upgrading:
...
+#!/bin/sh
+   
+/usr/sbin/logrotate -s /var/lib/logrotate/logrotate.status /etc/logrotate.conf
+EXITVALUE=$?
+if [ $EXITVALUE != 0 ]; then
+    /usr/bin/logger -t logrotate “ALERT exited abnormally with [$EXITVALUE]”
+fi
+exit 0
+```
+- We put the below script in /etc/cron.d so that logrotate check will be run every minute:
```
genwhittTTD, line 296: - We put the following script in /etc/cron.d so that the logrotate check is run every minute:

docs/guides/operator-guide-aws-marketplace.md
```diff
@@ -252,6 +252,69 @@ Here's what you need to know about upgrading:
...
+```
+# Run the minutely jobs
+SHELL=/bin/bash
+PATH=/sbin:/bin:/usr/sbin:/usr/bin
+MAILTO=root
+* * * * * root /usr/sbin/logrotate -s /var/lib/logrotate/logrotate.status /etc/logrotate.conf
+```
+- The reason why we have this as default setting is because:
```
genwhittTTD, line 304: Suggested mods for this line and the two below:

- This is the default setting for these reasons:
   - So that the maxsize condition is checked frequently
    - The command refers to /var/lib/logrotate/logrotate.status to check the log status and see if it has reached the daily rotation time, so that it won't make an extra rotation.

Note I'm not sure on the above... I don't entirely understand it, particularly the last bit "so it won't make an extra rotation"... just looking at the language, not the technicalities. So, your call.

docs/guides/operator-guide-aws-marketplace.md
```diff
@@ -252,6 +252,69 @@ Here's what you need to know about upgrading:
...
+    - We want maxsize condition to be checked minutely
+    - The command will refer /var/lib/logrotate/logrotate.status and check if it has reached the daily rotation time, so it won't make extra rotation
+
+### How to change log rotation schedule
```
genwhittTTD, line 308: ### How to change the log rotation schedule

docs/guides/operator-guide-aws-marketplace.md
```diff
@@ -252,6 +252,69 @@ Here's what you need to know about upgrading:
...
+Simply change the file etc/logrotate.d/uid2operator.conf following the logrotate documentation (https://linux.die.net/man/8/logrotate) will apply the custom change.
```
genwhittTTD, line 309: To change the log rotation schedule, update the `etc/logrotate.d/uid2operator.conf` file.

Follow the instructions in the logrotate documentation: see [logrotate(8) - Linux man page](https://linux.die.net/man/8/logrotate).

docs/guides/operator-guide-aws-marketplace.md
```diff
@@ -252,6 +252,69 @@ Here's what you need to know about upgrading:
...
+
+**The service does NOT need to be restarted to pick up the change.**
```
genwhittTTD, line 311: :::note
The service does NOT need to be restarted to pick up the change.
:::

docs/guides/operator-guide-aws-marketplace.md
```diff
@@ -252,6 +252,69 @@ Here's what you need to know about upgrading:
...
+
+### Useful commands
+- Run `sudo logrotate -f /etc/logrotate.conf --debug` to see detail explanation of what will be rotated
```
genwhittTTD, line 314: - To see a detailed explanation of what will be rotated, run `sudo logrotate -f /etc/logrotate.conf --debug`.

docs/guides/operator-guide-aws-marketplace.md
```diff
@@ -252,6 +252,69 @@ Here's what you need to know about upgrading:
...
+- Run `sudo logrotate -f /etc/logrotate.conf --force` to run one iteration of logrotate manually without scheduling it on some interval
```
genwhittTTD, line 315: - To run one iteration of logrotate manually without changing the scheduled interval, run `sudo logrotate -f /etc/logrotate.conf --force`.

docs/guides/operator-guide-aws-marketplace.md
```diff
@@ -252,6 +252,69 @@ Here's what you need to know about upgrading:
...
+
+### How to change log rotation schedule
+Simply change the file etc/logrotate.d/uid2operator.conf following the logrotate documentation (https://linux.die.net/man/8/logrotate) will apply the custom change.
+
+**The service does NOT need to be restarted to pick up the change.**
+
+### Useful commands
+- Run `sudo logrotate -f /etc/logrotate.conf --debug` to see detail explanation of what will be rotated
+- Run `sudo logrotate -f /etc/logrotate.conf --force` to run one iteration of logrotate manually without scheduling it on some interval 
+- To reload syslog-ng, run `sudo /usr/sbin/syslog-ng-ctl reload`
```
genwhittTTD, line 316: - To reload syslog-ng, run `sudo /usr/sbin/syslog-ng-ctl reload`.
(this is good, just add ending punctuation)

docs/guides/operator-guide-aws-marketplace.md
```diff
@@ -252,6 +252,69 @@ Here's what you need to know about upgrading:
...
+**The service does NOT need to be restarted to pick up the change.**
+
+### Useful commands
+- Run `sudo logrotate -f /etc/logrotate.conf --debug` to see detail explanation of what will be rotated
+- Run `sudo logrotate -f /etc/logrotate.conf --force` to run one iteration of logrotate manually without scheduling it on some interval 
+- To reload syslog-ng, run `sudo /usr/sbin/syslog-ng-ctl reload` 
+
 ## Technical Support
 
 If you have trouble subscribing or deploying the product, contact us at [aws-mktpl-uid@thetradedesk.com](mailto:aws-mktpl-uid@thetradedesk.com).
```
genwhittTTD, line 320: Curious where this email address came from. Just checking. Also, let's make it into a link so that the TTD email address isn't so obvious. Suggest:

If you have trouble subscribing to the product or deploying it, [contact us](mailto:aws-mktpl-uid@thetradedesk.com).

genwhittTTD (changes requested):
Quite a few wording mods. LMK if any questions. Happy to discuss and should definitely have another review.

docs/guides/operator-guide-aws-marketplace.md
cYKatherine: Checked with the team, Its a mailing list that goes to some of us who have related knowledge.

genwhittTTD (approved):

PR #597, opened 2024-06-05
wording updates re client-server, server-side

genwhittTTD:
- Fix instances server-side > client-server
- Fix instances server-only > server-side
- Publisher Integration Guide, Server-Only > Publisher Integration Guide, Server-Side
- Few other small minor edits


sunnywu (approved):

docs/guides/custom-publisher-integration.md
```diff
@@ -29,8 +29,8 @@ The following are the options available for publishers to integrate with UID2:
 
 - Client UID2 SDK for JavaScript (see [UID2 SDK for JavaScript Reference Guide](../sdks/client-side-identity.md)), with [UID2 SDK for Java](../sdks/uid2-sdk-ref-java.md) on the server.
 - Client UID2 SDK for JavaScript (see [UID2 SDK for JavaScript Reference Guide](../sdks/client-side-identity.md)), with custom server code.
```
sunnywu, line 31: your PR is fine but reading the above 2 lines are a bit confusing to me - i'm not sure what the above 2 means - is it saying the publisher needs to generate tokens on server-side using UID2 SDK for Java or custom server codes (i think it should be plural?) and then use JS SDK on the client-side to store the token and retrieve it for bidstream use. 

If this is the case, then shouldn't mentioning the Client-Server/Server-side integration below enough so we can remove the 2 lines above? 

the 2 lines above were from aug 2023 before we wrote up the client-server/server-side integration guides so i suspect the 2 lines above are just outdated now.


cc @jon8787 @lionell-pack-ttd for feedback.

lionell-pack-ttd: I agree - those first two options aren't needed any more.

docs/guides/custom-publisher-integration.md
```diff
@@ -29,8 +29,8 @@ The following are the options available for publishers to integrate with UID2:
...
-- Server-only integration, with [UID2 SDK for Java](../sdks/uid2-sdk-ref-java.md) or [UID2 SDK for Python](../sdks/uid2-sdk-ref-python.md) on the server.
-- Server-only integration, with custom server code.
+- Client-server integration, with [UID2 SDK for Java](../sdks/uid2-sdk-ref-java.md) or [UID2 SDK for Python](../sdks/uid2-sdk-ref-python.md) on the server.
+- Server-side integration, with custom server code.
 
 This guide provides information for the last two options.
```
lionell-pack-ttd, line 35: I think this guide now only provides information for the last option. If they're using a client-server integration (the second-last option), they'd be looking at the client-server guide.

docs/guides/custom-publisher-integration.md
sunnywu: ok sounds like we are in-sync here but see the comment below on what to edit

docs/guides/custom-publisher-integration.md
sunnywu: I think we can replace the whole paragraph with:

_See also [FAQs](http://localhost:3006/docs/guides/custom-publisher-integration#faqs).

This guide provides information for doing server-side integration using [UID2 SDK for Java](http://localhost:3006/docs/sdks/uid2-sdk-ref-java) or [UID2 SDK for Python](http://localhost:3006/docs/sdks/uid2-sdk-ref-python) or custom server codes, 

There is also an example application that demonstrates the workflow. See [Sample Application](http://localhost:3006/docs/guides/custom-publisher-integration#sample-application)._

PR #607, opened 2024-06-10
mods to add latency info to Environments and link others

genwhittTTD:
- Add general latency info to Environments file.
- Update other docs to link to that section but keep code samples in each file (Prebid x2, mobile x2, publisher client side, 5 in all)
- Updated mdx file, removed Japanese version.


docs/snippets/_sdk-reduce-latency.mdx
```diff
@@ -1,7 +1,5 @@
-By default, UID2 SDKs make API calls to a UID2 production environment server in the USA. Depending on where your users are based, you might consider choosing a server closer to your users to reduce latency.
+By default, the UID2 SDKs and Prebid.js implementation make API calls to a UID2 production environment server in the USA. Depending on where your users are based, you might consider choosing a server closer to your users to reduce latency.
```
jon8787, line 1: I think only the client-side SDKs (JS, prebid, android, ios) have a default URL. The server-side SDKs make you choose a URL (although this might change in future).
Would be better if there's a way to avoid the first sentence.

jon8787 (approved):

docs/snippets/_sdk-reduce-latency.mdx
genwhittTTD: Hi @jon8787: because this is a shared piece, it goes in both SDK and Prebid docs, so I included SDKs and Prebid in the copy. I appreciate it isn't all SDKs. Rather than getting even more specific, in the shared section I'm moving this content, which is implementation-specific, out of the shared piece to the individual documents.

docs/endpoints/post-identity-map.md
```diff
@@ -34,7 +34,7 @@ You must encrypt all requests using your secret. For details, and code examples
 
 | Path Parameter | Data Type | Attribute | Description |
 | :--- | :--- | :--- | :--- |
-| `{environment}` | string | Required | Integration environment: `https://operator-integ.uidapi.com`<br/>Production environment: `https://prod.uidapi.com`<br/>For a full list, including regional operators, see [Environments](../getting-started/gs-environments.md). |
+| `{environment}` | string | Required | Integration environment: `https://operator-integ.uidapi.com`<br/>Production environment: `https://prod.uidapi.com`<br/>For a full list of valid base URLs, including regional operators, see [Environments](../getting-started/gs-environments.md). |
```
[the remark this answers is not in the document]
jon8787: We do highlight that prod and integ are different in https://unifiedid.com/docs/getting-started/gs-environments
Do we really need to duplicate that information everywhere that we talk about URLs?

> But in this scenario is [prod.uidapi.com](https://prod.uidapi.com/) not a default because for the API there is no default, I guess that's what you're saying?
You're correct that direct API callers don't have any default. But I was referring to the fact that it's not appropriate to say:
"Production environment: https://prod.uidapi.com" in this document. Users will be misled to think that if they want production, that's the URL they should put. And that's not true, eg here's a quote from @atarassov-ttd :

> You would not want to hit [prod.uidapi.com](http://prod.uidapi.com/) for all advertisers:
> This increases latency
> Some advertisers may want to control which regions emails/phones are kept within, e.g. to comply with upcoming (?) Australia data protection laws some organisations want to make sure emails are never sent outside of Australia.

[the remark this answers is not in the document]
genwhittTTD: @jon8787 understood and thank you. Additional updates. Over to you.

docs/snippets/_sdk-reduce-latency.mdx
```diff
@@ -1,7 +1,5 @@
-By default, UID2 SDKs make API calls to a UID2 production environment server in the USA. Depending on where your users are based, you might consider choosing a server closer to your users to reduce latency.
+In this scenario, depending on where your users are based, you might consider choosing a server closer to your users to reduce latency.
```
jon8787, line 1: Not keen on talking about Singapore or global in this snippet, would rather than info was in the common Environments page

jon8787 (approved):

docs/snippets/_sdk-reduce-latency.mdx
genwhittTTD: @jon8787: updated. The detailed info is in the Environments page only, and other pages link to there. other than the implementation examples. Hope this is what you wanted. Snippet is unnecessary in this scenario, because of nuances of differences between pages.

PR #658, opened 2024-07-23
Add new instructions on how to scrape metrics for GCP

cYKatherine:


docs/guides/operator-private-gcp-confidential-space.md
```diff
@@ -500,7 +501,8 @@ import AttestFailure from '/docs/snippets/_private-operator-attest-failure.mdx';
 
 <AttestFailure />
 
-## Upgrading
+
+### Upgrading
```
cYKatherine, line 505: Change the title level because it should be wrapped under Task:
image

docs/guides/operator-private-gcp-confidential-space.md
```diff
@@ -339,7 +339,7 @@ To set up and configure the account that you created when you installed the gclo
     $ gcloud compute firewall-rules create operator-tcp \
       --direction=INGRESS --priority=1000 --network=default --action=ALLOW \
       --rules=tcp:8080 \
-      --source-ranges=0.0.0.0/0 \
+      --source-ranges={the ip address range for your service} \
```
cYKatherine, line 342: `{the ip range for your client}`

Also need to clarify what that mean

docs/guides/operator-private-gcp-confidential-space.md
```diff
@@ -509,12 +510,33 @@ If you're upgrading to a new version, the upgrade process depends on the deploym
...
+#### Upgrading&#8212;gcloud CLI
 
 If you deployed using the gcloud CLI, you must manually bring up new instances that use the new `{OPERATOR_IMAGE}` and then shut down old instances.
 
 If you previously set up a load balancer manually, you'll also need to update the mapping for the load balancer.
+
+## Scraping Metrics
+This section provides information on the process of scraping metrics. 
+
+We expose 9080 port which serves Prometheus metrics (/metrics).
```
cYKatherine, line 526: We provide prometheus format metrics on port 9080. You can use prometheus compatible scraper to collect and aggregate these metrics.

docs/guides/operator-private-gcp-confidential-space.md
```diff
@@ -339,9 +339,10 @@ To set up and configure the account that you created when you installed the gclo
     $ gcloud compute firewall-rules create operator-tcp \
       --direction=INGRESS --priority=1000 --network=default --action=ALLOW \
       --rules=tcp:8080 \
-      --source-ranges=0.0.0.0/0 \
+      --source-ranges=10.0.0.0/8 \
       --target-service-accounts={SERVICE_ACCOUNT_NAME}@{PROJECT_ID}.iam.gserviceaccount.com
     ```
+Note: The `source-ranges` should be the ip address for your clients. It is in the CIDR notaion.
```
thomasm-ttd, line 345: It is in CIDR notation - remove 'the'

docs/guides/operator-private-gcp-confidential-space.md
```diff
@@ -509,12 +511,17 @@ If you're upgrading to a new version, the upgrade process depends on the deploym
 - [Upgrading&#8212;Terraform Template](#upgradingterraform-template)
 - [Upgrading&#8212;gcloud CLI](#upgradinggcloud-cli)
 
-### Upgrading&#8212;Terraform Template
+#### Upgrading&#8212;Terraform Template
 
 If you deployed using the Terraform template, all you need to do to upgrade is update your deployment with the new `{OPERATOR_IMAGE}` that you received in the upgrade notification.
```
thomasm-ttd, line 516: I know this is not your change, but we need to change this. It should say:
If you deployed using the Terraform template, please compare the latest version of the template with the one you used to do the deploy.

docs/guides/operator-private-gcp-confidential-space.md
```diff
@@ -509,12 +511,17 @@ If you're upgrading to a new version, the upgrade process depends on the deploym
...
 
-### Upgrading&#8212;gcloud CLI
+#### Upgrading&#8212;gcloud CLI
 
 If you deployed using the gcloud CLI, you must manually bring up new instances that use the new `{OPERATOR_IMAGE}` and then shut down old instances.
 
 If you previously set up a load balancer manually, you'll also need to update the mapping for the load balancer.
+
+## Scraping Metrics
+This section provides information on the process of scraping metrics.
```
thomasm-ttd, line 525: I would remove this line

genwhittTTD (changes requested):
Few language suggestions and a couple of questions. Sequencing definitely needs some adjustment and we need to provide some upgrade direction for Terraform template.

docs/guides/operator-private-gcp-confidential-space.md
```diff
@@ -509,12 +511,15 @@ If you're upgrading to a new version, the upgrade process depends on the deploym
 - [Upgrading&#8212;Terraform Template](#upgradingterraform-template)
 - [Upgrading&#8212;gcloud CLI](#upgradinggcloud-cli)
 
-### Upgrading&#8212;Terraform Template
+#### Upgrading&#8212;Terraform Template
 
-If you deployed using the Terraform template, all you need to do to upgrade is update your deployment with the new `{OPERATOR_IMAGE}` that you received in the upgrade notification.
+If you have deployed private operator using the Terraform template, please compare the latest version of the template with the one you used to do the deploy.
```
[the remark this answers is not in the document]
cYKatherine: Changed to ```If you previously deployed the Private Operator for GCP using the Terraform template, compare the latest version of the template with the one you used when you deployed. If there are changes, make sure to redeploy to include all the updates.```

docs/guides/operator-private-gcp-confidential-space.md
cYKatherine: Good point, I think `Upgrading` should be at the same title level as `Tasks`. @thomasm-ttd Can you confirm this?

docs/guides/operator-private-gcp-confidential-space.md
```diff
@@ -339,9 +339,10 @@ To set up and configure the account that you created when you installed the gclo
     $ gcloud compute firewall-rules create operator-tcp \
       --direction=INGRESS --priority=1000 --network=default --action=ALLOW \
       --rules=tcp:8080 \
-      --source-ranges=0.0.0.0/0 \
+      --source-ranges=10.0.0.0/8 \
       --target-service-accounts={SERVICE_ACCOUNT_NAME}@{PROJECT_ID}.iam.gserviceaccount.com
     ```
+WARNING: The `source-ranges` specifies the range of ip addresses your clients will be making calls to the private operator from. It is in CIDR notaion, and can be comma-separated to provide multiple ranges.
```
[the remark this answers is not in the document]
cYKatherine: > 'source-ranges' is a -- what? Switch, parameter? If we say "the" we should say what type of thing it is.
`source-ranges` is the parameter we need when running the above command. If we quote it with the code formate (using `) developers should understand what it means.

> Re it being comma-separated, an example would be really nice. But maybe not necessary.
Added an example

> a) we should use Docusaurus admonition styling with the three colons before and after.
Changed

b) Not sure why it's a warning? Is there a risk? Suggest just a note.
Yes, the reason why we put `Warnings` is because the value of `source-range` is essential as it could potentially give access to unwanted ip addresses.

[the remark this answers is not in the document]
genwhittTTD: Hi @cYKatherine Thanks for all this info. 
Re "Yes, the reason why we put Warnings is because the value of source-range is essential as it could potentially give access to unwanted ip addresses."
Wondering if we should say something along those lines. For example: 

:::warning
`source-ranges` specifies the range of IP addresses from which your clients will call the Private Operator. It is in CIDR notation, and you can use comma-separated values to provide multiple ranges. Example: `--source-ranges="10.0.0.0/8,10.10.0.0/16"`. Make sure the ranges are accurate and include only IP addresses that belong to you.
:::

Notes: 1) in your update, WARNING is all caps which doesn't render correctly in Docusaurus. 2) Also you retained the old wording as well: you need to delete that. 3) Remove "the" before 'source-ranges' -- it's not grammatically accurate otherwise. 4) Suggest add last sentence as per the above.

docs/guides/operator-private-gcp-confidential-space.md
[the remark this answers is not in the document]
genwhittTTD: Thank you.

docs/guides/operator-private-gcp-confidential-space.md
thomasm-ttd: I think we should drop Tasks and the paragraph after that. 
So looks like this:
## Running the Health Check
### Private Operator Attestation Failure
## Upgrading
## Scraping Metrics

genwhittTTD (approved):
LGTM thx.

PR #664, opened 2024-08-02
UID2-2904 mentioned optout handling in JS/prebid.js CSTG guides

sunnywu:
1. In UID2-2904, we will return optout status instead of an optout token in CSTG endpoint to align the behaviour with EUID CSTG. SO updated both JS CSTG guide and prebid.js CSTG guide to mention it's possible to get optout status and may not always get a UID2 token.


docs/guides/integration-javascript-client-side.md
```diff
@@ -326,18 +338,20 @@ window.__uid2.callbacks.push(async (eventType, payload) => {
...
+          if (__uid2.hasOptedOut()) {
+            // DII has opted out of UID2, no UID2 token will be generated
+          }
+          else if (__uid2.isLoginRequired()) {
             // Call one of the setIdentityFrom functions to generate a new UID2 token.
             // Add any retry logic around this call as required.
             await __uid2.setIdentityFromEmailHash(
                 emailHash,
-                clientSideConfig
-          );
```
sunnywu, line 335: the indentations was wrong so actually there is syntax error. trying to fix it as well

docs/guides/integration-javascript-client-side.md
```diff
@@ -281,7 +281,19 @@ In some cases, the user's DII is not available on page load, and getting the DII
 You can potentially avoid that cost by checking for an existing token that you can use or refresh. To do this, call
 [__uid2.isLoginRequired](../sdks/sdk-ref-javascript#isloginrequired-boolean) which returns a Boolean value. If it returns `true`, this means that the UID2 SDK cannot create a new advertising token with the existing resource and DII is required to generate a brand new UID2 token.
 
-The following code snippet demonstrates how you might integrate with the UID2 SDK for JavaScript for the two scenarios above&#8212;starting with no token as well as reusing/refreshing any existing UID2 token if found. 
+The only exception is if the following method returns true,
```
sunnywu, line 284: alternatively we can just say:

_The only exception is the DII could be opted out of UID2 and no identity/token should be generated for it. The UID2 SDK for JavaScript will respect the opt out preference and not generate UID2 tokens even if you call any of the `setIdentity` method calls with DII again._

And don't mention the ```__uid2.hasOptedOut()``` method and leave it be.

docs/guides/integration-javascript-client-side.md
```diff
@@ -326,32 +338,39 @@ window.__uid2.callbacks.push(async (eventType, payload) => {
...
-        }
       }
       break;
  
     case "IdentityUpdated":
       // The IdentityUpdated event happens when a UID2 token is generated or refreshed.
       // See previous comment for an example of how the payload looks.
-      var advertising_token_to_use = payload.identity.advertising_token;
+      // It's possible that payload.identity is null because the DII has opted out of UID2
+      if (!payload.identity) {
```
sunnywu, line 362: if the user has opted out, the payload should not have a identity and we should add this condition to avoid misleading pubs to blindly get ```payload.identity.advertising_token```

If you agree, please check syntax tho

docs/guides/integration-prebid-client-side.md
```diff
@@ -144,6 +144,9 @@ pbjs.setConfig({
 }); 
 ```
 
+It is possible that the user's DII has opted out of UID2 previously. In this case, the UID2 module will respect the optout preference and no UID2 token will be generated and collected by Prebid.js.
```
sunnywu, line 147: i don't think we need to mention much here apart from saying prebid.js UID2 module will handle optout, unlike javascript sdk as pub needs to retrieve the ad token manually in latter

docs/guides/integration-javascript-client-side.md
```diff
@@ -326,32 +338,39 @@ window.__uid2.callbacks.push(async (eventType, payload) => {
         // }
         var advertising_token_to_use = payload.identity.advertising_token;
       } else {
-          if (__uid2.isLoginRequired()) {
+          if (__uid2.hasOptedOut()) {
```
sunnywu, line 341: is this possible?

docs/guides/integration-javascript-client-side.md
lionell-pack-ttd: I just tried this at https://euid.eu/docs/guides/integration-javascript-client-side using optout@example.com (which is opted out). I know this is EUID and not UID2 - I'm not sure if the behaviour is different. I'll try with our UID2 example site shortly.

image

It looks like nothing works as expected if the user has opted out :(

lionell-pack-ttd: I think isLoginRequired should return false if we have an optout identity. The user is logged in - they're opted out.

docs/guides/integration-javascript-client-side.md
lionell-pack-ttd: It could be null for other reasons as well - the user logged out, the token expired without being refreshed, maybe other things. And I'm not convinced the identity should be null if the user has logged out.

If they want an advertising token, they should check for it:
if (payload?.identity?.advertising_token) {

docs/guides/integration-javascript-client-side.md
```diff
@@ -281,7 +281,19 @@ In some cases, the user's DII is not available on page load, and getting the DII
...
+__uid2.hasOptedOut()
+```
+
+It indicates that the DII has been opted out of UID2 and no identity/token should be generated for it. While the UID2 SDK for JavaScript will respect the opt out preference and not generate UID2 tokens even if you call any of the `setIdentity` method calls with DII again, optionally, you might want to avoid making such calls repeatedly.
+
+The following code snippet demonstrates how you might integrate with the UID2 SDK for JavaScript for the three scenarios above:
+
+1. starting with no token 
+2. reusing/refreshing any existing UID2 token if found
+3. detect if the DII has opted out of UID2 and no token will ever be generated for it
```
lionell-pack-ttd, line 296: I think this phrasing is a little confusing - I think the original wording was clearer. I don't think we need to tell them that we're also handling the optout scenario here.

docs/guides/integration-javascript-client-side.md
sunnywu: ok adjusted as you suggested.

docs/guides/integration-javascript-client-side.md
sunnywu: ok reverted back to the original wording

docs/guides/integration-javascript-client-side.md
```diff
@@ -326,32 +328,37 @@ window.__uid2.callbacks.push(async (eventType, payload) => {
...
           }
-        }
       }
       break;
  
     case "IdentityUpdated":
       // The IdentityUpdated event happens when a UID2 token is generated or refreshed.
       // See previous comment for an example of how the payload looks.
-      var advertising_token_to_use = payload.identity.advertising_token;
+      // It's possible that payload/identity objects could be null for reasons such as user logged out, token
```
genwhittTTD, line 348: Are we sure we want to include "user logged out"? I removed a lot of mentions of login + logout, recently based on input from GM.

sunnywu: i think this particular instance is fine. can't think of a better way to word it

lionell-pack-ttd (approved):

PR #665, opened 2024-08-06
add new glossary term

genwhittTTD:
Add new glossary term, client keypair, plus adjust credentials text and add lead-in links.

### DIFF
--- a/docs/getting-started/gs-credentials.md
+++ b/docs/getting-started/gs-credentials.md
@@ -50,7 +50,7 @@ If you're using a client-side implementation (see [UID2 Client-Side Integration 
 
 **Client keypair** is a group term we use for these two values, which are used to uniquely define an account that's using an implementation that generates the token on the client side.
 
-When you're implementing UID2 on the client side, by using the UID2 JavaScript SDK, Client-Side Integration for Mobile, or Prebid, provide the values to the SDK or to Prebid as part of configuration.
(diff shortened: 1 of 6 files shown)


docs/getting-started/gs-credentials.md
```diff
@@ -48,7 +48,9 @@ If you're using a client-side implementation (see [UID2 Client-Side Integration
 - **Subscription ID**: This value identifies your site to the UID2 service.
 - **Public key**: This value is used for encryption.
 
-When you're implementing UID2 on the client side, by using the UID2 JavaScript SDK or Prebid, provide the values to the SDK or to Prebid as part of configuration.
+**Client keypair** is a group term we use for these two values, which are used to uniquely define an account that's using an implementation that generates the token on the client side.
+
+When you're implementing UID2 on the client side, by using the UID2 JavaScript SDK, Client-Side Integration for Mobile, or Prebid, provide the values to the SDK or to Prebid as part of configuration.
```
sunnywu, line 53: i think we can say Prebid.js instead of Prebid here - for using Prebid Mobile, they use keyapir when following Client-Side Integration for Mobile

sunnywu (approved):

docs/getting-started/gs-credentials.md
genwhittTTD: Thx. Fixed two instances on this line.

PR #692, opened 2024-08-29
Update ad token examples to v4

mcollins-ttd:
- Update ad token examples from v2 to v4
- Extract reusable snippets


genwhittTTD (approved):
LGTM! Can't answer for the accuracy of the token info but it builds perfectly and looks right :-) TYSM.

sunnywu (approved):

PR #694, opened 2024-08-29
UID2 token and refresh token new page, updated

genwhittTTD:
Add reference article with key information about UID2 token and refresh token.


docs/ref-info/ref-tokens.md
```diff
@@ -0,0 +1,56 @@
...
+
+- The UID2 token is a unique value: no two UID2 tokens are the same.
+- UID2 tokens are case sensitive.
+- The token value is opaque: do not make any assumptions about the format or about the length of the string.
+- The token has a limited life, but can be refreshed using the refresh token.
+- The token can be refreshed many times, to get a new UID2 token, as long as the UID2 token is always refreshed before the refresh token expires.
+- If the token has expired, or as an alternative to refreshing an existing token, you can always generate a new UID2 token from the original hashed or unhashed email address or phone number.
+- Publishers send UID2 tokens in the bidstream.
+- Refreshing a UID2 token does not invalidate/expire the original or previous UID2 token. You can still use the earlier token until it expires.
+- When the UID2 Operator service receives the refresh token with a request for a new UID2 token, it checks for user opt-out. If the user has opted out of UID2, no new UID2 token is generated. For details, see [User Opt-Out](../getting-started/gs-opt-out.md).
```
sunnywu, line 28: this line should also be copied and pasted in refresh token section below

Maybe in UID2 token section here, we say token generation logic will also check for user optout instead?

docs/ref-info/ref-tokens.md
```diff
@@ -0,0 +1,56 @@
...
+## UID2 Tokens: Key Information
+
+Here are some key points about UID2 tokens:
+
+- The UID2 token is a unique value: no two UID2 tokens are the same.
+- UID2 tokens are case sensitive.
+- The token value is opaque: do not make any assumptions about the format or about the length of the string.
+- The token has a limited life, but can be refreshed using the refresh token.
+- The token can be refreshed many times, to get a new UID2 token, as long as the UID2 token is always refreshed before the refresh token expires.
+- If the token has expired, or as an alternative to refreshing an existing token, you can always generate a new UID2 token from the original hashed or unhashed email address or phone number.
```
sunnywu, line 25: you can always generate a new UID2 token from the original hashed or unhashed email address or phone number _instead_. _Therefore it is optional to use the refresh token described below if you elect to just always generate new tokens from DII_

Make is clearer?

docs/ref-info/ref-tokens.md
```diff
@@ -0,0 +1,56 @@
...
+description: Information for publishers about UID2 tokens and refresh tokens.
+hide_table_of_contents: false
+sidebar_position: 06
+---
+
+import Link from '@docusaurus/Link';
+
+# UID2 Tokens and Refresh Tokens
+
+When a publisher sends a user's <Link href="../ref-info/glossary-uid#gl-dii">DII</Link> (email address or phone number) to the UID2 Operator, whether via one of the UID2 SDKs or the [POST&nbsp;/token/generate](../endpoints/post-token-generate.md) endpoint, the UID2 Operator returns a <a href="glossary-uid#gl-uid2-token">UID2 token</a> with associated values. The token is an opaque alphanumeric string, and is pseudonymous. Because of the way UID2 works, different instances of activity, on browsers, CTV, and electronic devices such as phone and tablets, can be matched without compromising the privacy of the individual. The token is designed so that it cannot be reverse engineered to arrive at the original email address or phone number.
```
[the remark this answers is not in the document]
genwhittTTD: @sunnywu thx much. Made the first change. Second one: yes, by associated values, which we use in several places in the doc, we mean the refresh token, timestamp etc. that come with the token.  I adjusted to specify refresh token as an example, but I think that wording is clear.
On the other points... this will go to Legal after you,. I can call out those sections but he will check those things anyway.

docs/ref-info/ref-tokens.md
genwhittTTD: Done on both... great edits, thx.

docs/ref-info/ref-tokens.md
```diff
@@ -9,21 +9,20 @@ import Link from '@docusaurus/Link';
...
+When a publisher sends a user's <Link href="../ref-info/glossary-uid#gl-dii">DII</Link> (email address or phone number) to the UID2 Operator, whether via one of the UID2 SDKs or the [POST&nbsp;/token/generate](../endpoints/post-token-generate.md) endpoint, the UID2 Operator returns a <a href="glossary-uid#gl-uid2-token">UID2 token</a> with associated values, including a refresh token. The token is an alphanumeric string. Each time a token is generated from DII input, the token value is different. Because of the way UID2 works, different instances of activity, on browsers, CTV, and electronic devices such as phone and tablets, can still be matched to the same <a href="glossary-uid#gl-raw-uid2">raw UID2</a> generated from the user's DII, even though the token might be different. The token is designed to protect against reverse engineering that might reveal the original email address or phone number.
```
sunnywu, line 12: i feel it is better to just copy the Glossary definition of UID2 token and put it here:

_UID2 tokens are generated from hashed or unhashed email addresses or phone numbers that are converted to raw UID2s and then encrypted._ 


---------------------------------------------------------------------------


_The token is designed to protect against reverse engineering that might reveal the original email address or phone number._

This makes it sound like people who can legit decrypt the token can know the original email/phone value (which isn't true as token can only be decrypted into a raw UID2. But yes if malicious party which has a mapping of raw UID2->original email/phone they can know the original email/phone.

We could reword
_The token is designed to protect the raw UID2 value inside the token so that only participant which have permission to decrypt can reveal the raw UID2 inside it._ 

Or just copy the glossary definition instead of reinventing the wheel here.
https://unifiedid.com/docs/ref-info/glossary-uid#gl-uid2-token

I feel current wording could be confusing and we should change it one way or another - i feel copy from glossary is easier.

docs/ref-info/ref-tokens.md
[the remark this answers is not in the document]
genwhittTTD: Follow up on this: these sections were changed.

docs/ref-info/ref-tokens.md
genwhittTTD: Majorly updated. from the glossary, to this:

When a publisher sends a user's DII—hashed or unhashed email addresses or phone numbers—to the UID2 Operator, whether via one of the UID2 SDKs or the POST /token/generate] endpoint, the UID2 Operator converts the DII to a raw UID2, encrypts it into a UID2 token, and returns the UID2 token with associated values, including a refresh token. The publisher can then use the UID2 token in the bidstream.

docs/ref-info/glossary-uid.md
```diff
@@ -307,6 +307,9 @@ import Link from '@docusaurus/Link';
 
 <dl>
 
+<dt><MdxJumpAnchor id="gl-opaque"><a href="#gl-opaque">Opaque</a></MdxJumpAnchor></dt>
+<dd>When we say a UID2 token is an opaque string, we mean that the way that the token is computed is not shared. No assumptions should be made about the format or length of the string, or any other aspect of it.</dd>
```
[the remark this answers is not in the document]
genwhittTTD: Thanks! Here's the latest per follow-up discussion also:

When we say a UID2 token is an opaque string, we mean that the way that the token is computed, and its format, are not communicated to UID2 participants and cannot be relied upon to remain unchanged. No assumptions should be made about the format or length of the string, or any other aspect of it.

sunnywu (approved):

PR #712, opened 2024-09-19
Prebid update for set-cookie info

genwhittTTD:
Prebid client-server doc:
- update Sample Token heading per JN
- update cookie example for realistic value


lionell-pack-ttd (approved):
LGTM!

PR #749, opened 2024-10-31
New portal features

alex-yau-ttd:
- Ability to switch participants
- User Roles
- Audit Trail


docs/portal/audit-trail.md
```diff
@@ -0,0 +1,12 @@
+---
+title: Audit Trail
+description: Set up and manage contacts for UID2 updates.
```
ashleysmithTTD, line 3: Looks like this is the same description from email contacts

docs/portal/team-members.md
```diff
@@ -18,17 +18,17 @@ In the UID2 Portal, on the **Team Members** page, you can perform all activities
...
-To access the Team Members page, go to the top right of the UID2 Portal user interface where your login ID is displayed. Click the down arrow and choose **Manage Team Members**.
+To access the Team Members page, go to the top right of the UID2 Portal user interface where your name is displayed. Click the down arrow and choose **Manage Team Members**.
 
 For an illustration, see [UID2 Portal Structure](portal-overview.md#uid2-portal-structure).
 
-## Team Member Rights and Responsibilities
+## User Roles
 
-Each team member is an administrator for the account, and can perform all activities relating to managing your team and your account, including the following:
+Each team member has one of two user roles: Admin or Operations. Admins can perform all activities related to UID2 participation, as summarized in the [UID2 Portal Overview](portal-overview.md).
```
ashleysmithTTD, line 27: nit: probably "Admin users" instead of "Admins" makes sense here for consistency

docs/portal/team-members.md
```diff
@@ -56,7 +60,7 @@ To re-send the invitation for a team member, complete the following steps:
 
 ## Edit Information for a Team Member
 
-You can update any of the values: First Name, Last Name, Email, or Job Function.
+Admins can update any of the values: First Name, Last Name, Email, Job Function, or Role.
```
ashleysmithTTD, line 63: nit: "Admin users" suggestion for here too

docs/portal/team-members.md
```diff
@@ -37,11 +37,15 @@ To add a team member, complete the following steps:
...
 1. At the top right, click the down arrow and choose **Manage Team Members**.
 1. Click **Add Team Member**.
-1. Enter the first name, last name, email, and job function.
+1. Enter the first name, last name, email, job function and role.
 1. Click **Save Team Member**.
 
    The team member is added to the list, with Pending status, and an invitation email is sent to the invited person.
 
+:::note
+Team members can be added to more than one participant.
```
ashleysmithTTD, line 46: "can belong" instead of "can be added".  I think here would be helpful even though it is implied to add a sentence that explicitly connects this sentence with the participant switcher - like "if a team member belongs to more than one participant, they will be able to switch between participants as shown in the image"

docs/portal/portal-overview.md
```diff
@@ -36,20 +37,23 @@ There are two main ways to access the pages, as shown in the following image:
 
 ![UID2 Portal Structure](images/portal-uid2-overview.png)
 
-From the left sidebar, you can access the following pages relating to configuring your account:
+Your participant name will be displayed at the top of the left sidebar. If your account has access to multiple participants, as shown in the example above, a dropdown menu will appear. You can use this menu to switch between participants in the portal.
```
ashleysmithTTD, line 40: Maybe add somewhere like on the Team Members page an explicit sentence that explains how an account does get access to a new participant outside of their default one - that an Admin for a participant needs to add someone as one of the team members for that person to get access.  Also this may be obvious, but when the admin adds them, they can add them with either Admin or Operations role to access to another participant - I'll let you make the call on that part.

ashleysmithTTD:
I noticed on the picture, the previous one has the name blacked out.  Not sure if its for legal reasons, but maybe worth checking with someone, or just blacking out the name on the new picture you uploaded

docs/portal/audit-trail.md
alex-yau-ttd: good catch - fixed!

docs/portal/team-members.md
alex-yau-ttd: yep good call. done.

docs/portal/team-members.md
alex-yau-ttd: yep good call. done.

docs/portal/team-members.md
alex-yau-ttd: I kept "can be added" because it is under the "Add a Team Member" section, but I've also made changes to use the "belong" terminology and connect it to the switcher.

docs/portal/portal-overview.md
alex-yau-ttd: I'm a bit hesitant to add details around how to invite an existing user to a new participant because it should just be the same as adding a brand new user to a participant. I'm planning on merging as-is and we can discuss/address separately.

alex-yau-ttd:
> I noticed on the picture, the previous one has the name blacked out. Not sure if its for legal reasons, but maybe worth checking with someone, or just blacking out the name on the new picture you uploaded

@ashleysmithTTD fair shout - I've just blacked out the name to err on the side of caution.

docs/portal/portal-overview.md
genwhittTTD: @alex-yau-ttd @ashleysmithTTD looking over these mods and I'd definitely be in favor of a) defining the roles in the doc, b) explaining how someone is assigned a role. It should be covered, I think.

PR #771, opened 2024-12-05
Gwh apidocs 2801 sample apps standardize uid2

genwhittTTD:
Standardize headings and wording for sample implementations referenced in the docs.

### DIFF
--- a/docs/guides/integration-google-ss.md
+++ b/docs/guides/integration-google-ss.md
@@ -111,14 +111,12 @@ A sample implementation is also available for integration using the SDK for Java
 
 The following sample implementations are available to illustrate how to integrate with the Google Ad Manager secure signals feature:
 
-- Client-Side UID2 SDK Integration Example:
-  - [Sample implementation](https://secure-signals-jssdk-integ.uidapi.com/)
-  - [Code repository](https://github.com/IABTechLab/uid2-web-integrations/tree/main/examples/google-secure-signals-integration/with_sdk_v3)
-- Client-Server UID2 SDK Integration Example:
+- Server-Side UID2 SDK Integration Example:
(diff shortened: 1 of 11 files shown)


docs/guides/integration-google-ss.md
```diff
@@ -105,20 +105,22 @@ This script is hosted on CDN, and GPT automatically loads it with the secure sig
...
-A sample application is also available for integration using the SDK for JavaScript. See [Sample Applications](#sample-applications).
+A sample implementation is also available for integration using the SDK for JavaScript. See [Sample Implementations](#sample-implementations).
 
-## Sample Applications
+## Sample Implementations
 
-The following sample applications are available to illustrate how to integrate with the Google Ad Manager secure signals feature:
+The following sample implementations are available to illustrate how to integrate with the Google Ad Manager secure signals feature:
 
-- Server-Side UID2 Integration Example:
```
sunnywu, line 114: these changes are not right -  Nothing here is "Client-side". I don't think this section requires any changes

So the original copy is actually correct to me. My understanding is:

Server-Side UID2 Integration Example:
- Generate token on server side 
- Only present and present the generated token on the webpage inside the browser and store the generated token inside cookie ad-hoc and DOES NOT use UID2 Javascript SDK inside the browser. 

Client-Server UID2 SDK Integration:
- Generate token on server side 
- store the generated token inside the UID2 Javascript SDK embedded on the webpage

The integration with the google secure signals are identical for both. 

@ssundahlTTD could you confirm if my understanding is correct?

genwhittTTD: Hi @sunnywu I've backed out this section pending input from @ssundahlTTD : merging the rest for now.

PR #779, opened 2024-12-11
Add missing optout_check

abuabraham-ttd:
Add the required optout_check 

https://unifiedid.com/docs/getting-started/gs-encryption-decryption

The code examples there for calling v2/token/generate doesn't have "optout_check:1" which we required for all new participants to add into the request since late last year.

 

We should specify like this:

echo '{"email": "sunny@example.com","optout_check":1}' | python3.10 uid2_request.py https://prod.uidapi.com/v2/token/generate .....


sunnywu (approved):

genwhittTTD (commented):
There are two instances in this file that also should probably  be changed ? Lines 116 and 812.

genwhittTTD (approved):

PR #803, opened 2025-01-13
Adding GCP and Azure doc

abuabraham-ttd:


docs/guides/operator-guide-azure-enclave.md
```diff
@@ -324,3 +324,21 @@ To upgrade, complete the following steps:
    ```
    for i in {0..COUNT}; az container delete --name uid-operator-OLD-VERSION-$i --resource-group {RESOURCE_GROUP} --yes
    ```
+
+## UID2 Operator Error Codes
+
+The following table lists errors that might occur during a Private Operator's startup sequence.
+
+:::note
+Error codes for Private Operator startup issues apply only to versions released in Q2 2025 and later.
```
genwhittTTD, line 333: This information was already pulled back: we should not include future error information in public documentation. Doc is for current products not future. Q2 2025 is very far in the future. Is there any current error information we can release?

docs/guides/operator-private-gcp-confidential-space.md
```diff
@@ -519,3 +519,20 @@ If you previously set up a load balancer manually, you'll also need to update th
 
 ## Scraping Metrics
 The Private Operator for GCP exposes [Prometheus-formatted metrics](https://prometheus.io/docs/concepts/data_model/) on port 9080 through the /metrics endpoint. You can use a Prometheus-compatible scraper to collect and aggregate these metrics for your own needs.
+
+## UID2 Operator Error Codes
+
+The following table lists errors that might occur during a Private Operator's startup sequence.
+
+:::note
+Error codes for Private Operator startup issues apply only to versions released in Q2 2025 and later.
```
genwhittTTD, line 528: This information was already pulled back: we should not include future error information in public documentation. Doc is for current products not future. Q2 2025 is very far in the future. Is there any current error information we can release?

genwhittTTD (changes requested):
See comments. Not sure why this is coming back. Maybe I'm missing some data.

docs/guides/operator-guide-azure-enclave.md
```diff
@@ -324,3 +324,21 @@ To upgrade, complete the following steps:
...
+:::
+
+| Error Code | Issue | Steps to Resolve |
```
abuabraham-ttd, line 336: would be good to add a small section that says, if operator.json is modified the cce policy will change and containers wont start

PR #806, opened 2025-01-27
ibg-TTD-786-tables

ibgrav:
- Add styles to html table headers to match markdown table headers


genwhittTTD (approved):
Loogs great! TYSM.

PR #849, opened 2025-03-12
add server-side integration guide for mobile

genwhittTTD:
add server-side integration guide for mobile + in-links in other files.


docs/guides/integration-mobile-overview.md
```diff
@@ -53,6 +53,7 @@ For details, refer to one of the following guides:
 
 - [Client-Side Integration Guide for Mobile](integration-mobile-client-side.md)
 - [Client-Server Integration Guide for Mobile](integration-mobile-client-server.md)
+- [Server-Side Integration Guide for Mobile](integration-mobile-server-side.md)
```
caroline-ttd, line 56: Nitpick, In "Client-Side or Client-Server Integration?" section, probably we can change to "Client-Side, Client-Server or Server-side Integration?" and server side scenario in the table

caroline-ttd: Add also "Do you want to retrieve the UID2 token on the client side or on the server side" to "Do you want to retrieve the UID2 token on the client side, on the client-server side or on the server side?"

caroline-ttd: Hi Gen, thank you, yeah, that makes sense!

In the L81, do we want to add the scenario in the table?
something like "You have access to DII on the server side only and want to manage the UID2 token entirely on the server side"

genwhittTTD: Yes -- TY! I adjusted the client-server line slightly also. See update.
Note I cannot make these copy mods in the Japanese version -- translator will do that.

caroline-ttd (approved):

PR #853, opened 2025-03-14
feat: :sparkles: add perion partner

bbuilds:
Added Perion partner as DSP.

Screenshot 2025-03-14 at 1 50 44 PM


genwhittTTD (approved):
All good thx.

PR #869, opened 2025-04-04
minor mods to publisher docs

genwhittTTD:
minor mods to publisher docs:
* Add Private Operator info into a couple of the publisher docs.

### DIFF
--- a/docs/snippets/_private-operator-option.mdx
+++ b/docs/snippets/_private-operator-option.mdx
@@ -1,0 +1,12 @@
+import Link from '@docusaurus/Link';
+
+If you're a publisher and are concerned about keeping your first-party data (emails or telephone numbers) within your corporate network, you might choose to deploy a <Link href="../ref-info/glossary-uid#gl-private-operator">Private Operator</Link>. In this scenario, you choose to host a private instance exclusively for your own use.
+
+For additional information, review these resources:
+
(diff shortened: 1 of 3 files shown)


docs/snippets/_private-operator-option.mdx
```diff
@@ -0,0 +1,12 @@
+import Link from '@docusaurus/Link';
+
+If you're a publisher and are concerned about keeping your first-party data (emails or telephone numbers) within your corporate network, you might choose to deploy a <Link href="../ref-info/glossary-uid#gl-private-operator">Private Operator</Link>. In this scenario, you choose to host a private instance exclusively for your own use.
```
sunnywu, line 3: why do we say "emails or telephone numbers" not DII? it's fine to me but it's a little inconsistent

sunnywu: _If you're a publisher and are concerned about having your first-party data leaving your corporate network_ 

or 

_If you're a publisher and wants to keep your first-party data within your corporate network_ 

both are better than the current draft i think.

sunnywu (approved):

docs/snippets/_private-operator-option.mdx
genwhittTTD: Thx Sunny. Went with the second option.

PR #895, opened 2025-06-26
Jpn translation update 881 893

miyaichi:
Pull requests #881 to #893 (excluding #886 and #8897) have been reflected in the Japanese documentation.


genwhittTTD (approved):
Thanks!

PR #902, opened 2025-07-01
UID2-5741 Add notes about --kubernetes-version

cYKatherine:


docs/guides/operator-guide-aks-enclave.md
```diff
@@ -277,6 +277,9 @@ az aks create \
     --nodepool-name oprnodepool \
     --os-sku Ubuntu
 ```
+:::note
+Be sure to use the latest supported Kubernetes version, using the --kubernetes-version flag. If you use an earlier version, you must enable Long Term Support (LTS). For details, see Long-term support for [Azure Kubernetes Service (AKS) versions](https://learn.microsoft.com/en-us/azure/aks/long-term-support).
```
mcollins-ttd, line 281: `--kubernetes-version` should be formatted as code.

mcollins-ttd: I think the link text should include "Long-term support for".

caroline-ttd (approved):

mcollins-ttd (approved):

genwhittTTD (changes requested):
Looks good! One edit marked otherwise good to go.
Note: I was going to suggest we could make these changes in the Japanese file as well, but there's a mix of code changes and copy changes so I agree with your choice to leave those for the translator :-)

PR #904, opened 2025-07-03
Identity Map V3 Java SDK Doc

aulme:
- Tests for code samples are here: https://github.com/IABTechLab/uid2-client-java/pull/96
- Some of the existing code samples were incorrect, fixed those too.


docs/sdks/sdk-ref-java.md
```diff
@@ -302,6 +306,175 @@ If you're using server-side integration (see [Publisher Integration Guide, Serve
 
 ## Usage for Advertisers/Data Providers
 
+1. Create an IdentityMapV3Client as an instance variable:
+   ```java
+   final private IdentityMapV3Client identityMapV3Client = new IdentityMapV3Client(UID2_BASE_URL, UID2_API_KEY, UID2_SECRET_KEY);
+   ```
+
+2. Create an IdentityMapV3Input object. You can use emails, phone numbers, or both:
```
sophia-chen-ttd, line 314: should we include that they can also use their hashed forms?

PR #929, opened 2025-08-06
Update docs for Mobile SDK v2 releases

dcaunt:
Update docs ahead of Android and iOS 2.0.0 releases, which only contains breaking changes to the `UID2Prebid` module. 

The main change is in the following two files where we detail version numbers for Prebid integration:
- docs/guides/integration-mobile-client-server.md
- docs/guides/integration-mobile-client-side.md

Most other changes are simply to reference version 2.0.0 as the latest version.

### DIFF
--- a/docs/sdks/sdk-ref-ios.md
+++ b/docs/sdks/sdk-ref-ios.md
@@ -51,9 +51,9 @@ The steps you'll take in the UID2 Portal are different depending on whether your
 
 ## SDK Version
 
-<!-- As of 22 Oct 2024 -->
+<!-- As of 2025-08-07 -->
 
-This documentation is for the SDK for iOS version 1.7.0 or later.
+This documentation is for the SDK for iOS version 2.0.0 or later.
 
 For current and past release notes information, see [https://github.com/IABTechLab/uid2-ios-sdk/releases](https://github.com/IABTechLab/uid2-ios-sdk/releases).
 
@@ -90,7 +90,7 @@ Add the following dependency to Package.swift:
 
 ```js
 dependencies: [
-  .package(url: "https://github.com/IABTechLab/uid2-ios-sdk.git", from: "1.7.0"),
+  .package(url: "https://github.com/IABTechLab/uid2-ios-sdk.git", from: "2.0.0"),
 ]
 ```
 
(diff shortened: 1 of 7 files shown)


docs/snippets/_mobile_docs_prebid-mobile.mdx
```diff
@@ -32,20 +32,20 @@ To configure your UID2 Prebid for Mobile integration, follow these steps:
...
    ```js
-   pod 'UID2Prebid', '~> 1.7'
+   pod 'UID2Prebid', '~> 2.0'
    ```
    </TabItem>
    <TabItem value='spm' label='SPM (iOS)'>
-   :::warning
-   Integration via Swift Package Manager is not supported until it is supported by the Prebid Mobile SDK itself. For details, see [Prebid's Mobile SDK documentation](https://docs.prebid.org/prebid-mobile/pbm-api/ios/code-integration-ios.html#swift-pm).
-   :::
+   ```js
```
genwhittTTD, line 46: Could we have a line of instruction above the code, as we have for the other flavors? Something along these lines:

* Gradle:   Include the following in your Gradle configuration:
* Maven:    To install with Maven, add the SDK as a dependency in the `pom.xml` file:
* CocoaPods:   Add the following entry in your `Podfile`:

docs/sdks/sdk-ref-ios.md
```diff
@@ -100,14 +100,14 @@ In the XCode user interface, under Package Dependencies, add the following entry
 
 | Name | Location | Dependency Rule |
 | :--- | :--- | :--- |
-| uid2-ios-sdk | `git@github.com:IABTechLab/uid2-ios-sdk.git` | Up to next major version: 1.7.0 < 2.0.0 |
+| uid2-ios-sdk | `git@github.com:IABTechLab/uid2-ios-sdk.git` | Up to next major version: 2.0.0 < 2.0.0 |
```
genwhittTTD, line 103: This might be fine -- I just don't get it. A smaller version to a greater version number I understand, but both values are the same. Does the wording need to be updated?

Up to next major version: 2.0.0 < 2.0.0

genwhittTTD (changes requested):
Two small comments otherwise great. Thx!

genwhittTTD (approved):
LGTM, thx!

PR #944, opened 2025-10-07
Handle cases for opted out emails

eiman-eltigani-ttd:
Currently when users enter an opted out email in the prebid sample site, nothing happens. Updated functionality so user knows whats happening similar to how EUID prebid sample sites displays an opted out message.

### DIFF
--- a/static/examples/cstg-prebid-example/index.html
+++ b/static/examples/cstg-prebid-example/index.html
@@ -64,9 +64,6 @@
 
       function onDocumentReady() {
         console.log('Setting up interface handlers.');
-        // Clear storage on page load for fresh start
-        localStorage.removeItem('__uid2_advertising_token');
-        
         $('#login').click(handleLogin);
         $('#clear_storage').click(handleClearStorage);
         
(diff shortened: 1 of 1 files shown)


static/examples/cstg-prebid-example/index.html
```diff
@@ -10,22 +10,42 @@
...
+        const storedToken = localStorage.getItem('__uid2_advertising_token');
+        if (storedToken) {
+          try {
+            const tokenData = JSON.parse(storedToken);
+            isOptedOut = tokenData.latestToken === 'optout';
+          } catch (e) {
+            console.log('Could not parse stored token');
+          }
+        }
+
```
eiman-eltigani-ttd, line 33: Unlike EUID which returns object with "optout" field when email is opted out, UID2 pbjs just returns null so we have to check in local storage whether the user is opted out.

static/examples/cstg-prebid-example/index.html
```diff
@@ -97,7 +117,6 @@
       var pbjs = pbjs || {};
       pbjs.que = pbjs.que || [];
       pbjs.que.push(function () {
-        setUid2Config();
```
eiman-eltigani-ttd, line 100: Removed to prevent initialziation conflicts + keep consistent with EUID

PR #1009, opened 2026-04-15
Add inbound/outbound rule for Azure

cYKatherine:


jon8787 (dismissed):

docs/guides/operator-guide-azure-enclave.md
```diff
@@ -179,6 +179,18 @@ The following diagram illustrates the virtual private cloud that hosts a UID2 Pr
 
 ![VPC Network](images/operator-azure-drawio.png)
 
+#### Network Security Group Policy
+
+:::note
+To avoid passing certificates associated with your domain into the enclave, inbound HTTP is allowed instead of HTTPS. This also avoids the cost of a secure layer, if used in a private network that is internal to your organization.
```
genwhittTTD: We say inbound HTTP is allowed but really I think we mean inbound HTTPS is NOT allowed. 

Also one thing I'm not sure about technically. We say it avoids the extra cost... IF used in a private network. Presumably it's always a private network. The "if" implies it might not be... in which case what would you do if HTTPS is not allowed? So I think it's not quite clear. LMK if I'm misunderstanding.

My stab at rewording:


```suggestion
To avoid passing certificates associated with your domain into the enclave, only inbound HTTP is allowed. Inbound HTTPS is not allowed. This also avoids the extra cost of another secure layer in a network that's already private and internal to your organization.
```

docs/guides/operator-guide-azure-enclave.md
```diff
@@ -179,6 +179,18 @@ The following diagram illustrates the virtual private cloud that hosts a UID2 Pr
...
+:::
+
```
genwhittTTD: We have a section heading, then a note, then the table. We need some intro text to the table, per our style guide. Suggest (but please clarify as needed, I might not be understanding fully):


```suggestion
The following table provides information about supported protocols.
```

docs/guides/operator-guide-azure-enclave.md
```diff
@@ -179,6 +179,18 @@ The following diagram illustrates the virtual private cloud that hosts a UID2 Pr
...
+| Port Number | Direction | Protocol | Description |
+| ----------- | --------- | -------- | ------ |
+| 80 | Inbound | HTTP | Serves all UID2 APIs, including the healthcheck endpoint `/ops/healthcheck`.<br/>When everything is up and running, the endpoint returns HTTP 200 with a response body of `OK`. For details, see [Running the Health Check](#running-the-health-check). |
```
genwhittTTD: healthcheck > health check

We use two words in UID2 docs (other than the actual endpoint name of course).

docs/guides/operator-guide-azure-enclave.md
```diff
@@ -179,6 +179,18 @@ The following diagram illustrates the virtual private cloud that hosts a UID2 Pr
 
 ![VPC Network](images/operator-azure-drawio.png)
 
+#### Network Security Group Policy
```
genwhittTTD: @cYKatherine a key concern is the placement of this new content. In the existing doc there is a heading, https://unifiedid.com/docs/guides/operator-guide-azure-enclave#set-up-the-vpc-network, Set Up the VPC Network, and there is a diagram followed by steps. This new section has been added between the diagram and the steps, which is not correct.

In the AWS doc, which was mentioned in the thread (https://unifiedid.com/docs/guides/operator-guide-aws-marketplace#vpc-chart), it's right before the VPC chart, and at the same heading level. Therefore, for consistency, I think you should move this whole new section to above the "Set Up the VPC Network" heading (line 174 in the existing doc) and made it an h3 (three hashtags) rather than what you have which is an h4. That will be more parallel with the AWS doc.

genwhittTTD (commented):
Some copy edits plus please move the section to above the diagram and make it an h3. Thx much.

genwhittTTD (changes requested):
Hi @cYKatherine . No edits but looking at the output, sorry but I think this is still in the wrong place. It's a separate section which is good... but it's reference information, and it's in the middle of a set of procedure sections.

Looking at the current file, I think that it should be with the intro sections that give people the information they need before they actually get started. I think it should go after this section: https://unifiedid.com/docs/guides/operator-guide-azure-enclave#deployment-environments. And be an h2.

I was trying to make it parallel with AWS but, looking at it, AWS has it in the Prerequisites section which isn't really correct because (I think) it isn't a specific prerequisite, it's reference data, though it's important. But let's not mess with that for now.

If you don't think my suggestion is right, my other suggestion is to put it nearer the end and link to it. But I think the above is best. LMK what you think.

All else is good.

genwhittTTD (approved):
@cYKatherine absolutely fine. Just checking if we should not cross-reference from anywhere earlier to help people see this info. But the new section is fine where you have it now, and LGTM -- approving. TYSM.

PR #1012, opened 2026-04-17
UID2-6917 Add 20 concurrent requests restriction for /identity/map

swibi-ttd:
- Add our suggested limit of 20 concurrent requests for `/identity/map` endpoints
- Fix Documentation Updates entry that was in the wrong quarter

### DIFF
--- a/docs/endpoints/post-identity-map-v2.md
+++ b/docs/endpoints/post-identity-map-v2.md
@@ -28,7 +28,7 @@ If you're using an earlier version, we recommend that you upgrade as soon as pos
 Here's what you need to know:
 
 - The maximum request size is 1MB.
-- To map a large number of email addresses, phone numbers, or their respective hashes, send them in batches with a maximum of 5,000 items per batch. You may send up to 20 batches in parallel.
+- To map a large number of email addresses, phone numbers, or their respective hashes, send them in batches with a maximum of 5,000 items per batch. We recommend sending no more than 20 batches in parallel.
 - Be sure to store mappings of email addresses, phone numbers, or their respective hashes.<br/>Not storing mappings could increase processing time drastically when you have to map millions of email addresses or phone numbers. Recalculating only those mappings that actually need to be updated, however, reduces the total processing time because only about 1/365th of raw UID2s need to be updated daily. See also [Advertiser/Data Provider Integration Overview](../guides/integration-advertiser-dataprovider-overview.md) and [FAQs for Advertisers and Data Providers](../getting-started/gs-faqs.md#faqs-for-advertisers-and-data-providers).
 
 ## Rate Limiting
--- a/docs/endpoints/post-identity-map-v2.md
+++ b/docs/endpoints/post-identity-map-v2.md
@@ -23,12 +23,12 @@ This documentation is for version 2 of this endpoint, which is not the latest ve
 If you're using an earlier version, we recommend that you upgrade as soon as possible, to take advantage of improvements. For migration guidance, see [Migration from v2 Identity Map](post-identity-map.md#migration-from-v2-identity-map). For deprecation information, see [Deprecation Schedule: Endpoint Versions](../ref-info/deprecation-schedule.md#endpoint-versions).
 :::
 
-## Batch Size Requirements
+## Batch Size and Request Parallelization Requirements
 
 Here's what you need to know:
 
 - The maximum request size is 1MB.
-- To map a large number of email addresses, phone numbers, or their respective hashes, send them in batches with a maximum of 5,000 items per batch.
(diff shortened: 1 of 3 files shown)


cYKatherine (dismissed):

docs/endpoints/post-identity-map-v2.md
```diff
@@ -23,12 +23,12 @@ This documentation is for version 2 of this endpoint, which is not the latest ve
...
 :::
 
-## Batch Size Requirements
+## Batch Size and Request Parallelization Requirements
 
 Here's what you need to know:
 
 - The maximum request size is 1MB.
-- To map a large number of email addresses, phone numbers, or their respective hashes, send them in batches with a maximum of 5,000 items per batch.
+- To map a large number of email addresses, phone numbers, or their respective hashes, send them in batches with a maximum of 5,000 items per batch. You may send up to 20 batches in parallel.
```
genwhittTTD: Say "can" rather than "may" here. We're not giving them permission, we're stating capabilities. (though per the description this is a recommendation? Either way it's not quite right. But we should be clear whether it's a recommendation or a requirement. Suggest:

```suggestion
- To map a large number of email addresses, phone numbers, or their respective hashes, send them in batches with a maximum of 5,000 items per batch. We recommend sending no more than 20 batches in parallel.
```

Or if it's an even more firm recommendation, we could say:

Do not send more than 20 batches in parallel.

That also leaves me wondering why, though. If there's anything we can comfortably share it would be good. Very likely not but checking in case we could say something like this:

To avoid processing delays, do not send more than 20 batches in parallel.

That gives them a purpose. Though, again, it might be too much info.

docs/endpoints/post-identity-map.md
```diff
@@ -23,12 +23,12 @@ This documentation is for the latest version of this endpoint, version 3.
...
 If needed, documentation is also available for the previous version: see [POST /identity/map (v2)](post-identity-map-v2.md).
 
-## Batch Size Requirements
+## Batch Size and Request Parallelization Requirements
 
 Here's what you need to know:
 
 - The maximum request size is 1MB.
-- To map a large number of email addresses, phone numbers, or their respective hashes, send them in batches with a maximum of 5,000 items per batch.
+- To map a large number of email addresses, phone numbers, or their respective hashes, send them in batches with a maximum of 5,000 items per batch. You may send up to 20 batches in parallel.
```
genwhittTTD: Same here.

docs/ref-info/updates-doc.md
```diff
@@ -20,9 +20,25 @@ Check out the latest updates to our UID2 documentation resources.
 Use the Tags toolbar to view a subset of documentation updates.
 :::
 
-## Q1 2026
+## Q2 2026
```
genwhittTTD, line 23: We only announce major documents, basically... not small fixes like this. Suggest just remove the new section (leave the fix of course since it's a correction to existing content). Though actually my comment applies to the Rate Limiting changes also.

If there's some specific reason we should include it, OK... but there have been many, many changes that we don't announce. I looked over the doc and appreciate that this isn't clear from the heading. 

This is something I'll ask SW and KK about. I'll copy you on it. For now, your call.

docs/ref-info/updates-doc.md
```diff
@@ -20,9 +20,25 @@ Check out the latest updates to our UID2 documentation resources.
...
+
+<CustomTagsContainer tags="Endpoints">
+
+### Request Parallelization Updates for POST /identity/map
+
+April 17, 2026
+
+The [POST&nbsp;/identity/map](../endpoints/post-identity-map.md) endpoint documentation has been updated with the following changes:
+
+- Added a recommendation to send up to 20 /identity/map batches in parallel.
```
genwhittTTD: Just saying in case you end up keeping it -- the way this is written it doesn't sound like a limit. It sounds like we're recommending sending lots. Suggest use the same rewording as on the other entries, if that makes sense.

Added a recommendation to send not more than 20 batches in parallel.

Something like that. But better we don't list a doc change this small.

genwhittTTD (changes requested):
Thx Sean. Few comments. Also I will open a thread about the level of detail in Doc Updates and copy you on it.

docs/ref-info/updates-doc.md
swibi-ttd: Good point - I didn't think about how Doc Updates would get cluttered if we add all small changes. I've now removed the entry. Anyways, this change had already been communicated to existing participants so the changelog wouldn't be particularly useful to them

genwhittTTD (approved):
LGTM thx!

cYKatherine (approved):

PR #1023, opened 2026-05-22
Updates to some doc titles to remove UID2 prefix

genwhittTTD:
Updates to some doc titles to remove UID2 prefix (plus corresponding updates to links to doc titles, front matter, and sidebar)


i18n/ja/docusaurus-plugin-content-docs/current/intro.md
```diff
@@ -53,7 +53,7 @@ UID2 インフラストラクチャの主要な要素や、UID2 のその他の
 - [UID2 identifier types](ref-info/uid-identifier-types.md)
 - [UID2 components](ref-info/uid-components.md)
 - [UID2 participants](overviews/participants-overview.md)
-- [Workflows](ref-info/uid-workflows.md)
+- [UID2 workflows](ref-info/uid-workflows.md)
```
justinleonTTD, line 56: Everything else was removing UID2 but this one and the sidebars.js are adding it. Just want to make sure that was intentional.

genwhittTTD: @justinleonTTD great spot but yes it was intentional but in the doc titles only  not on the sidebar, per input from KK. They are not guide titles but are a bit short if the UID2 is removed... but on the sidebar we have UID2/EUID as the parent and the single words in the child positions. But then when you click through it says UID2 (or EUID).

All links to these docs should be consistent with that, and between English and Japanese and between UID2 and EUID (and I believe they are).

justinleonTTD (approved):

ericwTTD (approved):

PR #1025, opened 2026-06-01
sidebar cleanup UID2

genwhittTTD:
Cleanup mods to the exception sidebars. Also updated the caniuse browser list which was out of date.


sidebars.js
```diff
@@ -426,9 +426,6 @@ const sidebars = {
     'guides/integration-publisher-server-side',
     'GAM Secure Signals',
     'guides/integration-google-ss',
-    'Mobile',
-    'guides/sdk-ref-android',
```
nathanieltjandraTTD, line 430: Should this be /sdks/sdk-ref-android instead of removing it? Same for case for /sdks/sdk-ref-ios.

Repeat for all.

nathanieltjandraTTD (commented):
Left comments.

sidebars.js
genwhittTTD: @Nathaniel Tjandra I thought so at first but on examination, no. See notes in the Jira.

In the SDKs section, all participants should have access to all SDK docs -- per the table of audiences that we did at the time: https://thetradedesk.atlassian.net/wiki/spaces/UID2/pages/25234491/UID2+Doc+Pages+Per+Audience (look at the entries for the SDKs left-menu item: all are labelled for all audiences).

Where these entries appeared was in the Guides section, which is probably where the erroneous folder name came from. Mobile is a section in the Publisher docs: https://unifiedid.com/docs/guides/integration-mobile-overview. Not sure, but I think re history that we put the SDK docs in that section before the guides were available.

I'm sure it's correct as is. The Publisher section comes out right, and so do all the others.

nathanieltjandraTTD (approved):

PR #1026, opened 2026-06-05
chore: upgrade GitHub Actions from Node 20 to Node 24

mcollins-ttd:
Node 20 is being deprecated on GitHub Actions runners. Starting **June 16, 2026**, runners will use Node 24 by default.

This PR upgrades outdated action references in this repo to the latest released version that ships Node 24 support, pinned to its commit SHA. Each action's node version is determined by reading its `action.yml` at the pinned ref.

See: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -24,8 +24,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
         with:
           node-version: 20
           cache: npm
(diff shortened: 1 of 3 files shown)


.github/workflows/deploy.yml
```diff
@@ -1,44 +1,44 @@
...
+          node-version: 20
+          cache: npm
+      - name: Install dependencies
+        run: npm ci
+      - name: Build website
+        run: npm run build
+      - name: Setup Pages
+        uses: actions/configure-pages@45bfe0192ca1faeb007ade9deae92b16b8254a0d # v6.0.0
+      - name: Upload artifact
+        uses: actions/upload-pages-artifact@v3
```
mcollins-ttd: **Claude Review:** `actions/upload-pages-artifact@v3` was not upgraded here. v3 uses Node 20, and unlike the other four actions in this file, it was left on a floating tag rather than pinned to a commit SHA. `upload-pages-artifact@v4` uses Node 24 — you can pin it the same way as the others: `actions/upload-pages-artifact@<sha> # v4.x.x`.

sophia-chen-ttd (approved):

PR #1032, opened 2026-06-16
UID2-7002 Add Private Operator Network Egress page with list of domains

swibi-ttd:
Add a new page Private Operator Network Egress listing the outbound destinations a Private Operator must reach, so participants with restrictive network requirements can self-serve instead of requesting the list ad-hoc.


sidebars.js
```diff
@@ -259,6 +259,7 @@ const fullSidebar = [
             'guides/operator-private-gcp-confidential-space',
             'guides/operator-guide-azure-enclave',
             'guides/operator-guide-aks-enclave',
+            'guides/operator-private-network-requirements',
```
genwhittTTD: @swibi-ttd I'd prefer if this goes in the ref-info folder, because it's really reference info not a guide. The only thing we have in the guides folder, other than guides, is the overview summaries for same.

The guides refer to many other docs with various types of reference info but they're in the ref-info folder.

This would also of course mean updating the links.

docs/ref-info/operator-private-network-requirements.md
```diff
@@ -0,0 +1,39 @@
+---
```
genwhittTTD, line 1: As per my comment on the sidebars.js file: I'd much prefer if this is in the ref-info folder -- more logical for the type of info.

Also: you'll need to copy the file into the Japanese fileset. Otherwise, there are broken out-links at runtime. It's a Docusaurus bug I think. Just until the translator catches up with actually translating the content.

You'll see if you build for both languages (npm run build && npm run serve)

docs/guides/operator-private-network-requirements.md
```diff
@@ -0,0 +1,39 @@
...
+| Hostname | Purpose |
+| --- | --- |
+| `core-prod.uidapi.com` | Core Service (attestation, keys, salts, configuration) |
+| `optout-prod.uidapi.com` | Opt-Out Service |
+| `uid2-core-prod-store.s3.us-east-2.amazonaws.com` | Core data storage |
+| `uid2-core-prod-store-replica.s3.us-west-2.amazonaws.com` | Core data storage (failover replica) |
+| `uid2-optout-prod-store.s3.us-east-2.amazonaws.com` | Opt-out data storage |
+| `uid2-optout-prod-store-replica.s3.us-west-2.amazonaws.com` | Opt-out data storage (failover replica) |
+
+Allow these by hostname rather than by IP address, as the underlying addresses change.
```
genwhittTTD: "as" has multiple meanings so better to use "because.

Also, probably they might change rather than they do change? If changing is a facet of what they are, "because the underlying addresses change" is fine.

```suggestion
Allow these by hostname rather than by IP address, because the underlying addresses might change.
```

genwhittTTD (changes requested):
@swibi-ttd great work. Couple of copy comments. Mostly, I'd like this to live in the Reference Info folder.

Also, can't remember if I noted this -- you'll need to copy the English file to the corresponding Japanese location until we get it translated. Docusaurus is smart enough to pick up the English file in the Japanese, but out-links are broken.

sidebars.js
swibi-ttd: Makes sense! I've moved it to `ref-info` and updated links.

docs/ref-info/operator-private-network-requirements.md
swibi-ttd: Copied the file to the Japanese fileset. Works now when I build for both languages.

swibi-ttd:
@genwhittTTD Thank you for your feedback! In addition I've also made the following changes:
- Renamed the page to `Private Operator network egress` to be more technically accurate. Also updated links accordingly.
- Added a link to the new page at the bottom of the parent `Private Operator integration overview` page, as per your suggestion in Slack

Changes have also been reflected in the EUID PR.

docs/ref-info/operator-private-network-requirements.md
```diff
@@ -0,0 +1,39 @@
...
+import Link from '@docusaurus/Link';
+
+# Private Operator network egress
+
+A <Link href="../ref-info/glossary-uid#gl-private-operator">Private Operator</Link> connects to the UID2 <Link href="../ref-info/glossary-uid#gl-core-service">Core</Link> and <Link href="../ref-info/glossary-uid#gl-opt-out-service">Opt-Out</Link> services, and downloads data files directly from AWS S3 using URLs that the Core service provides. For details, see [Private Operator workflow](../guides/integration-options-private-operator.md#private-operator-workflow).
+
+If your environment restricts outbound network traffic, you must allow outbound HTTPS (port 443) to all of the destinations below, or the operator cannot start.
+
+## Integration
+
```
genwhittTTD: Sorry, Sean, I missed one thing. We need a lead-in line before the tables (our style standards). For both tables. For example:


```suggestion
The following table lists the hostnames you must allow for the integration environment.
```

docs/ref-info/operator-private-network-requirements.md
```diff
@@ -0,0 +1,39 @@
...
+| Hostname | Purpose |
+| --- | --- |
+| `core-integ.uidapi.com` | Core Service (attestation, keys, salts, configuration) |
+| `optout-integ.uidapi.com` | Opt-Out Service |
+| `uid2-core-integ-store.s3.us-east-2.amazonaws.com` | Core data storage |
+| `uid2-optout-integ-store.s3.us-east-2.amazonaws.com` | Opt-out data storage |
+
+## Production
+
```
genwhittTTD: Same comment as previous.

I'm going to approve with comments so that you can fix and merge, so you're not held up.

But, **remember to update this file and then copy the latest to the Japanese. Easy to forget that!** Here's recommended wording:


```suggestion
The following table lists the hostnames you must allow for the production environment.
```

genwhittTTD (dismissed):
I forgot to ask for intros to the two new tables. And then copy the updated file to the Japanese filesystem.

Approving so you can merge, but please make these mods. TYSM.

sidebars.js
genwhittTTD: Perfect, thx!

genwhittTTD (approved):
LGTM thx.

PR #1037, opened 2026-06-26
UID2-7375 Reference UID Verify from publisher web integration guides

sunnywu:
## Summary

UID Verify (the Chrome debugging extension) was previously documented only on its standalone reference page ([`ref-info/ref-uid-verify`](https://unifiedid.com/docs/ref-info/ref-uid-verify)). It was not referenced from the publisher integration guides.

This PR adds a shared snippet (`docs/snippets/_snpt-uid-verify-debug-tip.mdx`) with a `:::tip` pointing to the UID Verify reference page, framed as a debugging/inspection aid (not a token-generation integration option), and imports it into:

**Hub pages:**
- `docs/snippets/_snpt-publisher-implementation-resources.mdx` — renders on [Publisher integration resources](https://unifiedid.com/docs/guides/integration-options-publisher-all) and the Publishers overview
- `docs/guides/integration-options-publisher-web.md` — new "Debug your integration" section

**Individual web guides** (the three integration types UID Verify supports):
- `integration-javascript-client-side.md`
- `integration-javascript-client-server.md`
- `integration-prebid-client-side.md`
- `integration-prebid-client-server.md`
- `integration-google-ss.md`

The tip text lives in a single snippet so it stays consistent across all pages. Additions only; no existing content changed. Pure server-side guide is excluded (nothing client-side for UID Verify to inspect). Japanese i18n mirror left for the docs team to translate separately.

## Jira
[UID2-7375](https://thetradedesk.atlassian.net/browse/UID2-7375)

## Related
EUID-docs companion PR: https://github.com/European-Unified-ID/EUID-docs/pull/425

🤖 Generated with [Claude Code](https://claude.com/claude-code)


docs/guides/integration-google-ss.md
```diff
@@ -20,6 +21,8 @@ This guide covers integration steps for publishers using UID2 with the Google Ad
 To use the UID2 Google Ad Manager secure signals integration, if you are using an SDK you must have your UID2 integration already set up. This does not apply if you are using server-side integration. For a summary of all the integration options available, see [Integration guides: Summary](summary-guides.md).
 :::
 
+<SnptUidVerifyDebugTip />
```
genwhittTTD: visual issue here -- a Note immediately followed by a Tip, better not to do.

Suggest have the content here without the Tip tag?

Another possibility is to add it as an item in the Troubleshooting article (https://unifiedid.com/docs/guides/integration-google-ss#troubleshooting).

Or we could change the preceding Note into a prerequisites section.

docs/guides/integration-javascript-client-side.md
```diff
@@ -37,6 +38,8 @@ UID2 provides an SDK for JavaScript (see [SDK for JavaScript reference guide](..
 - Automatic refreshing of UID2 tokens
 - Automatic storage of UID2 tokens in the browser
 
+<SnptUidVerifyDebugTip />
```
genwhittTTD: I don't love the placement here -- between the SDK features and the summary of steps to implement, which I think flow well one to the other. Perhaps it could go after the summary of steps? But keep it here if you prefer.

docs/guides/integration-javascript-client-server.md
```diff
@@ -43,6 +44,8 @@ For integration scenarios for publishers that do not use the SDK for JavaScript,
 If you are using Google Ad Manager and want to use the secure signals feature, first follow the steps in this guide and then follow the additional steps in the [Google Ad Manager Secure Signals integration guide](integration-google-ss.md).
 :::
 
+<SnptUidVerifyDebugTip />
```
genwhittTTD: Again, here, it's added in right below another admonition. We avoid consecutive admonitions if possible. And in this case there is even another one before that, with only one line of copy between. So we have Tip, one line, Note, then this new Tip.

I had an idea. What about having a separate section directly after the "Integrating with single sign-on (SSO)" section in applicable guides, and calling it "Using UID Verify Chrome Extension"? That would mean consistent placement in all applicable docs.

docs/guides/integration-options-publisher-web.md
```diff
@@ -138,3 +139,7 @@ The following table shows integration options that support passing UID2 token in
 :::note
 As long as you generate the token and keep it refreshed, you can also use other options for passing the UID2 token into the bidstream.
 :::
+
+## Debug your integration
```
genwhittTTD: Here it's a separate section, which I think is good.

How about we call it "Using UID Verify Chrome Extension"? Or, "Inspecting with UID Verify"? Or some other name you're comfortable with for the instances in the other guides so that in all instances it could be a separate section. Yours is better, actually, because it says what it does. But right at the top of the other docs is too soon to introduce debugging info.

Whatever you settle on for the title, remove the "tip" tag. We should not have a whole section with only a tip in it and no other copy.

docs/snippets/_snpt-publisher-implementation-resources.mdx
```diff
@@ -27,6 +28,8 @@ For a detailed summary of web integration options, see [Publisher web integratio
 | Server-side integration | [Publisher integration guide, server-side](../guides/integration-publisher-server-side.md) | A guide for publishers who do not use the [SDK for JavaScript](../sdks/sdk-ref-javascript.md). |
 | Publisher/SSP integration with GAM | [Google Ad Manager Secure Signals integration guide](../guides/integration-google-ss.md) | A guide covering the additional steps needed for publishers using UID2 with the Google Ad Manager Secure Signals feature (previously known as Encrypted Signals for Publishers, ESP). |
 
+<SnptUidVerifyDebugTip />
```
genwhittTTD: _snpt-publisher-implementation-resources is used in a couple of overview-type docs including the main Publishers overview/intro page. I think that it's better not on those pages. It's debug info. Better to just have it as a heading on the pages for the specific docs is my thought, with the separate heading as suggested in earlier comments (and no Tip tag).

genwhittTTD (changes requested):
Left some comments. The tip copy itself is great but I'm suggesting different placement -- a separate section, below the "Integrating with single sign-on (SSO)" section which is in many docs -- no Tip tag, don't include it in the main intro and overview pages. Those are my suggestions.

docs/guides/integration-google-ss.md
sunnywu: change note to prerequisites is better - Adding a Troubleshooting section at the bottom may be overlooked.

docs/guides/integration-options-publisher-web.md
sunnywu: Yeh i think "Inspecting with UID Verify Chrome Extension" is good.

docs/snippets/_snpt-publisher-implementation-resources.mdx
sunnywu: sounds good

docs/guides/integration-options-publisher-web.md
genwhittTTD: I made these mods. FYI I changed the capitalization... UID Verify is a product name,  Chrome is a product name, but extension is not. So, with sentence capitalization, Extension > extension.

docs/guides/integration-google-ss.md
genwhittTTD: @sunnywu : I added in the "Inspecting with UID Verify Chrome extension" section just after the SSO section, as I'd suggested, for all 5 integration guides.

Also I've added the SSO section into the publisher web options page since it wasn't there... LMK if you disagree with that..

In case you were talking about the Note in the intro section. I started to do this but it's already covered in the first step, "Complete UID2 account setup and configure account" ... and adding a Prerequisites section is more work. So, just focusing on getting these UID Verify sections in the right places.

docs/guides/integration-javascript-client-server.md
genwhittTTD: Done as per other comments. It's a separate section, in 6 files.

docs/guides/integration-javascript-client-side.md
genwhittTTD: Updated so it's after the SSO section, consistently with all affected files.

genwhittTTD (approved):
Approving my own changes :-) But over to you for review, @sunnywu

swibi-ttd (approved):

